PR #13, opened 2020-02-23
Set up CI with Azure Pipelines

mratsim:


mratsim:
Last failure on 32-bit should be fixed by commit https://github.com/mratsim/constantine/pull/14/commits/550b7b1660c62ddbef58ace139dd4f916f46baf0 which fixed ARM compilation in Travis CI branch

Issue #15, opened 2020-02-23
Implement lazy carries and reductions


mratsim:
## Context

Currently after each addition or substraction steps there is a reduction done if the result is over the field modulus.

Due to constant-time constraints, there is no shortcut if it is unnecessary, the memory accesses are always done.

Instead, at the cost of a couple bits, we could use lazy carries/reductions.

Instead of using 31 bits over 32-bit words or 63-bit over 64-bit word, we use less bits in each words. For example assuming we use 26 bits and 52 bits respectively from 32 or 64-bit words, we only need to reduce every 6 or 12 additions respectively.
This is desirable in particular for [addition chains](https://en.wikipedia.org/wiki/Addition_chain)

## Choosing a default: Maximizing the usage of the word overhead

256-bit curves are quite common:
- secp256k1 for blockchain ECDSA / transaction signing
- P256/secp256r1 or Curve25519 for Diffie-Hellman

Or closely 254-bit Barreto-Naehrig for zkSNARKS.

Representing a 256 or 254 bits curve in the most compact way on 64-bit arch requires 4 words. Assuming
- we want at most a 1 word overhead per 256-bit
- we want to maximize the laziness of reduction

52-bit logical words and 12 lazy bits or 51-bit logical words and 13 lazy bits would be the best logical word bitsizes. They both can represent 2^255 integers in 5 words (but a radix 2^52 representation can also represent the numbers 2^255 and 2^256 in 5 words)

## Side-note on SIMD

This may also enable opportunities for SIMD vectorization using either integer or floating-point math.

- 2^24+1 is the first integer that cannot be represented in float32 though leaving 8-bit on the table might arguably be too much.
- 2^53+1 is the first integer that cannot be represented in float64
- The new AVX512_IFMA instructions (Integer Fused-Multiply-Add) supports multiply-add of 52-bit integers: `VPMADD52LUQ` and `VPMADD52HUQ`

Using floating point for pairings is covered in this paper:

-  New software speed records for cryptographic pairings\
  Michael Naehrig, Ruben Niederhagen, and Peter Schwabe\
  https://cryptojedi.org/papers/dclxvi-20100714.pdf

- Pairings on elliptic curves -- parameter selection and efficient computation\
  Michael Naehrig\
https://cryptosith.org/michael/data/talks/2010-10-19-ECC.pdf

## Side-note on "final substraction"-less Montgomery Multiplication/Exponentiation

[...]

PR #17, opened 2020-03-15
Internals refactor + renewed focus on perf

mratsim:
This PR is a complete overhaul of Constantine internal.

## Change of priorities

This stems for a renewed focus on performance.
Instead of focusing on constant-time, code size and performance in this order, the library will focus on
constant-time, performance and code size.

The new focus on performance is due to the following articles https://medium.com/loopring-protocol/zksnark-prover-optimizations-3e9a3e5578c0, https://hackmd.io/@zkteam/goff and the [Cambrian explosion of Zero-Knowledge Proof protocols](https://nakamoto.com/cambrian-explosion-of-crypto-proofs/) (ZKP).

In particular the first post shows that machines used for ZKP protocols start from $1000 (16 cores + 64GB of RAM) and in discussion at EthCC with Consensys ZKP team, I realised that clusters with ~100 of cores would be interesting to use.
At those scale, squeezing the most performance possible from the low-level implementation would significantly reduce the cost of the hardware, and might even make assembly (and it's auditing) worthwhile in the future.

## Performance

Here are the performance figures, before/after.

### GCC abysmal performance

Note that GCC generates very inefficient and also bloated code for multiprecision arithmetic, even when using addcarry and subborrow intrinsics.
This is so bad that GMP has a dedicated web page: https://gmplib.org/manual/Assembly-Carry-Propagation.html

Example in Godbolt: https://gcc.godbolt.org/z/2h768y
```C
#include <stdint.h>
#include <x86intrin.h>

void add256(uint64_t a[4], uint64_t b[4]){
  uint8_t carry = 0;
  for (int i = 0; i < 4; ++i)
    carry = _addcarry_u64(carry, a[i], b[i], &a[i]);
}
```

GCC
```asm
add256:
        movq    (%rsi), %rax
        addq    (%rdi), %rax
        setc    %dl
        movq    %rax, (%rdi)
        movq    8(%rdi), %rax
        addb    $-1, %dl
        adcq    8(%rsi), %rax
        setc    %dl
        movq    %rax, 8(%rdi)
        movq    16(%rdi), %rax
        addb    $-1, %dl
        adcq    16(%rsi), %rax
        setc    %dl
        movq    %rax, 16(%rdi)
        movq    24(%rsi), %rax
        addb    $-1, %dl
        adcq    %rax, 24(%rdi)
        ret
```

Clang
```asm
add256:
        movq    (%rsi), %rax
        addq    %rax, (%rdi)
        movq    8(%rsi), %rax
        adcq    %rax, 8(%rdi)
        movq    16(%rsi), %rax
        adcq    %rax, 16(%rdi)
        movq    24(%rsi), %rax
        adcq    %rax, 24(%rdi)
        retq
```

There are a couple of issues related in GCC tracker:
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67317 (fixed but fix is wrong?)
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79173 (support add carry and sub borrow)
- https://gcc.gnu.org/legacy-ml/gcc-help/2017-08/msg00100.html (GCC is not able to distinguish carry chains)

### Benchmark & Compilation flags

Nim devel from Jan 13 2020
-d:danger

Benchmark is https://github.com/mratsim/constantine/blob/191bb771/benchmarks/bench_eth_curves.nim
which benchmarks the library on the 3 Ethereum 1 and 2 elliptic curves:
- secp256k1 (Ethereum1 ECDSA)
- BN254 (Ethereum 1 precompile and the Zero-Knowledge Proof standard curve for Zcash and many many others)
- BLS12_381 (Ethereum 2 signatures and standard across Algorand, Chia, Dfinity, Ethereum 2, Filecoin, ...)

The most important item is the field Multiplication it's the building block that makes everything (exponentiation and inversion in particular) slow.

**Important: My CPU is overclocked, the hardware clock is using the CPU nominal frequency instead of the overclocked frequency meaning the benchmark are only meaningful to compare between runs on my own PC**

### GCC before PR

```
$  build/bench_eth_curves_gcc_old


Warmup: 0.9042 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[Secp256k1]          13 ns        39 cycles
Substraction    Fp[Secp256k1]           8 ns        26 cycles
Negation        Fp[Secp256k1]           3 ns        11 cycles
Multiplication  Fp[Secp256k1]          59 ns       179 cycles
Squaring        Fp[Secp256k1]          59 ns       179 cycles
Inversion       Fp[Secp256k1]       23215 ns     69646 cycles


Warmup: 0.8972 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BN254]              13 ns        39 cycles
Substraction    Fp[BN254]               8 ns        26 cycles
Negation        Fp[BN254]               4 ns        12 cycles
Multiplication  Fp[BN254]              59 ns       179 cycles
Squaring        Fp[BN254]              59 ns       179 cycles
Inversion       Fp[BN254]           23049 ns     69149 cycles


Warmup: 0.8966 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BLS12_381]          17 ns        51 cycles
Substraction    Fp[BLS12_381]          10 ns        32 cycles
Negation        Fp[BLS12_381]           4 ns        14 cycles
Multiplication  Fp[BLS12_381]         106 ns       320 cycles
Squaring        Fp[BLS12_381]         106 ns       319 cycles
Inversion       Fp[BLS12_381]       62882 ns    188649 cycles
```

### Clang before PR

```
$  build/bench_eth_curves_clang_old


Warmup: 0.9157 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[Secp256k1]          12 ns        37 cycles
Substraction    Fp[Secp256k1]           8 ns        24 cycles
Negation        Fp[Secp256k1]           4 ns        14 cycles
Multiplication  Fp[Secp256k1]          55 ns       167 cycles
Squaring        Fp[Secp256k1]          55 ns       167 cycles
Inversion       Fp[Secp256k1]       20619 ns     61860 cycles


Warmup: 0.9060 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BN254]              12 ns        37 cycles
Substraction    Fp[BN254]               8 ns        24 cycles
Negation        Fp[BN254]               4 ns        12 cycles
Multiplication  Fp[BN254]              55 ns       167 cycles
Squaring        Fp[BN254]              55 ns       167 cycles
Inversion       Fp[BN254]           20555 ns     61666 cycles


Warmup: 0.9054 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BLS12_381]          16 ns        49 cycles
Substraction    Fp[BLS12_381]          10 ns        31 cycles
Negation        Fp[BLS12_381]           4 ns        14 cycles
Multiplication  Fp[BLS12_381]         101 ns       304 cycles
Squaring        Fp[BLS12_381]         101 ns       304 cycles
Inversion       Fp[BLS12_381]       54204 ns    162615 cycles
```

### GCC after PR
```
$  build/bench_eth_curves_gcc_new


Warmup: 0.9033 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[Secp256k1]           4 ns        14 cycles
Substraction    Fp[Secp256k1]           3 ns        10 cycles
Negation        Fp[Secp256k1]           2 ns         6 cycles
Multiplication  Fp[Secp256k1]          34 ns       104 cycles
Squaring        Fp[Secp256k1]          34 ns       104 cycles
Inversion       Fp[Secp256k1]       12463 ns     37390 cycles


Warmup: 0.8966 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BN254]               4 ns        13 cycles
Substraction    Fp[BN254]               3 ns        10 cycles
Negation        Fp[BN254]               2 ns         6 cycles
Multiplication  Fp[BN254]              32 ns        98 cycles
Squaring        Fp[BN254]              32 ns        98 cycles
Inversion       Fp[BN254]           11473 ns     34420 cycles


Warmup: 0.8966 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BLS12_381]           9 ns        27 cycles
Substraction    Fp[BLS12_381]           5 ns        15 cycles
Negation        Fp[BLS12_381]           3 ns        10 cycles
Multiplication  Fp[BLS12_381]          62 ns       188 cycles
Squaring        Fp[BLS12_381]          62 ns       188 cycles
Inversion       Fp[BLS12_381]       31324 ns     93972 cycles
```

### Clang after PR

```
$  build/bench_eth_curves_clang_new


Warmup: 0.9139 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[Secp256k1]           3 ns         9 cycles
Substraction    Fp[Secp256k1]           2 ns         7 cycles
Negation        Fp[Secp256k1]           0 ns         0 cycles
Multiplication  Fp[Secp256k1]          22 ns        68 cycles
Squaring        Fp[Secp256k1]          22 ns        68 cycles
Inversion       Fp[Secp256k1]        9779 ns     29339 cycles


Warmup: 0.9064 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BN254]               2 ns         8 cycles
Substraction    Fp[BN254]               2 ns         6 cycles
Negation        Fp[BN254]               0 ns         0 cycles
Multiplication  Fp[BN254]              21 ns        64 cycles
Squaring        Fp[BN254]              21 ns        64 cycles
Inversion       Fp[BN254]            9264 ns     27794 cycles


Warmup: 0.9052 s, result 224 (displayed to avoid compiler optimizing warmup away)


⚠️ Measurements are approximate and use the CPU nominal clock: Turbo-Boost and overclocking will skew them.
==========================================================================================================

Addition        Fp[BLS12_381]           3 ns        11 cycles
Substraction    Fp[BLS12_381]           2 ns         8 cycles
Negation        Fp[BLS12_381]           0 ns         0 cycles
Multiplication  Fp[BLS12_381]          45 ns       136 cycles
Squaring        Fp[BLS12_381]          45 ns       137 cycles
Inversion       Fp[BLS12_381]       24951 ns     74855 cycles
```

## Code-size

The code-size increase to support 3 curves (with BN254 and secp256k1 using the same number of limbs) is very reasonable over the old code.
image

## Explanation of the internals

The old internals were using the same representation as BearSSL [BigInt](https://bearssl.org/bigint.html), in particular for words uint64 words only 63-bit were used and the last one stored the carry flag as there is no easy access to the carry flag in C.
The issue caused by carries is visible in this code to handle carries in Nim compile-time VM:
https://github.com/mratsim/constantine/blob/3c9d07003b18142586f355097c0b2a4bad88e90b/constantine/arithmetic/precomputed.nim#L34-L95
vs the old representation: https://github.com/mratsim/constantine/blob/191bb7710c2c5c378ef1a65b7316206af867b6cf/constantine/arithmetic/precomputed.nim#L37-L55

However the BearSSL representation has a couple of issues:
- It uses more words, a 254-bit field like for BN254 or a 381-bit field like for BLS12_381 requires an extra word compared to a compact representation. On current CPUs, the biggest performance bottleneck is memory speed, we want to access as little memory as possible, and for multiprecision multiplication those accesses are quadratic.
- The BigInt primitives are implemented via type-erasure / pointer indirections, this is great for code size but it prevents unrolling and inlining and an addition on 4 limbs is only 8 instructions (4 mov, and "add+adc+adc+adc" chain) which should be unrolled and inline.
Lastly at least with Clang we have access to efficient add with carry intrinsics.

So the new representation uses the full 64-bit and uses intrinsics or uint128 to deal with add with carries.

Furthermore it uses the technique described in https://hackmd.io/@zkteam/modular_multiplication to improve speed while staying at a high-level.

## Why no lazy carries or reduction

As mentioned https://github.com/mratsim/constantine/issues/15 lazy carries and reductions seem to be popular. Those also have issues:

- They significantly increase the number of memory accesses and so cache misses. At worse ADC latency is 6 cycles, a cache miss is 100 cycles. Arguably everything is stack allocated and always in cache so this argument might not hold but it also increases register pressure.
- Addition chains can have lazy carries but
  - when a substraction is involved either the representation is signed to handle lazy substraction
    or substraction requires to reduce the Field Element
  - Multiplications require a reduction (even partial, iirc below 2p is fine)
  - Unless we have a prime of special form (Generalized Mersenne Prime or Golden Prime) reduction is very costly as we can shift the carries from one limb to the other in a single pass, multiple (constant-time) conditional substractions and inequality checks will be needed.
  - This makes auditing the library harder

### CMOV / Ccopy

Currently conditional mov and copy use assembly and do a `test` before.
The test only needs to be done once when looping over bigints.
Thankfully the impact should be very small or invisible due to instruction level parallelism.

### Squaring

This is planned

### Multiplication

Further speed improvements are possible but will probably require either inflexible Assembly/inline assembly (i.e. always compiled-in and with predetermined number of limbs) or a mini-compiler.
An example for multi-precision addition is available in https://github.com/mratsim/constantine/blob/cc75582f9e618ea4304dd6e1bd7bc73d79034467/constantine/primitives/research/addcarry_subborrow_compiler.nim#L34-L79
As explained in A Fast Implementation of the Optimal Ate Pairing over BN curve on Intel Haswell Processor, Shigeo Mitsunari, 2013, https://eprint.iacr.org/2013/362.pdf one of the main bottleneck on x86 is that the MUL instruction is very inflexible in terms of register and requires lots of mov before and after which significantly hinder throughput of Montgomery multiplication.
Furthermore, it pollutes the carry flags, by using MULX instead we can avoid that and even use ADCX and ADOX instructions to handle 2 independent carry chains and benefit from instruction-level parallelism as mentioned by Intel in https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/ia-large-integer-arithmetic-paper.pdf

image
image


mratsim:
Comparing:
- Ubuntu 16.04 failure x86_64 GCC (Travis): https://travis-ci.com/github/mratsim/constantine/jobs/298511157#L354
- Ubuntu 16.04 failure x86_64 Clang (Travis): https://travis-ci.com/github/mratsim/constantine/jobs/298511159#L352
- Ubuntu 16.04 failure x86_64 Clang (Travis): https://travis-ci.com/github/mratsim/constantine/jobs/298511159#L352
- Ubuntu 16.04 failure GCC Azure: https://dev.azure.com/numforge/Constantine/_build/results?buildId=387&view=results
With
- MacOS x86_64 Clang success: https://travis-ci.com/github/mratsim/constantine/jobs/298511160#L247 (overall run failure is due to log size)
- Windows and Mac success (Azure): https://dev.azure.com/numforge/Constantine/_build/results?buildId=387&view=results

The issue doesn't seem to be in moveMem.

Could it be an issue in the carry flags and ccopy interaction at 
https://github.com/mratsim/constantine/blob/25972a280bbc081db08fee1385bf0a653e743e49/constantine/arithmetic/limbs.nim#L258-L286

mratsim:
Fixed the 2 leftover bugs:
1. GCC before version 7 generated wrong code on addcarry_u64
2. The fallback for subborrow_u64 on ARM didn't mask the borrow byte.

PR #20, opened 2020-04-07
Nim v1.2.0 update

mratsim:


mratsim:
1.2.0 not working on Linux might be related to the corruption from https://github.com/nim-lang/Nim/issues/13902 as `Word` is a `distinct uint64`

mratsim:
So Nim 1.2.2 has other limiting bugs related to static:
image
/home/vsts/work/1/s/constantine/elliptic/ec_endomorphism_accel.nim(307, 9) Error: invalid type for const: static[Curve](BLS12_381)

https://github.com/mratsim/constantine/blob/5591d17bdfd605cfa6d581a8b37cbaad9f6bd882/constantine/elliptic/ec_endomorphism_accel.nim#L297-L310

Nim 1.2.x is probably not meant to be. Closing the PR.

PR #28, opened 2020-06-03
Scalar mul tests

mratsim:
This introduces:
- scalar multiplication tests and fuzzing against a reference double-and-add implementation
- elliptic cure test vector generation via a Sagemath script
  - [x] BN254
  - [x] BLS12-381
- Fixes to the scalar multiplication implementation


mratsim:
The square root failure on FKM12_447 will be investigated in https://github.com/mratsim/constantine/issues/30

PR #29, opened 2020-06-04
Github Action CI (without GMP)

mratsim:
Initial Github Action trigger.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       max-parallel: 20
       matrix:
-        branch: [version-1-2, devel]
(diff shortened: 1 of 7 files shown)


mratsim:
Note like #20 we need to wait for https://github.com/nim-lang/Nim/issues/13902 to be released to stable (via Nim v1.2.2) to merge the tests that uses Nim stable

mratsim:
Failures on Windows 32-bit:

image

On Windows 64-bit

image

mratsim:
Linux 32-bit inconsistent operant in ASM
image

mratsim:
For now leaving windows aside. I'd have to SSH into Github CI to start debugging the windows bugs. They don't happen on Azure even though both are running on the same Microsoft backend (?).

Issue #53, opened 2020-06-18
Trigger edge-cases with randomized testing


mratsim:
Constantine is currently using randomized testing vs a reference implementation (GMP) and property-based testing which uncovered edge cases (tagged heisenbugs):
- https://github.com/mratsim/constantine/issues/30
- https://github.com/mratsim/constantine/issues/42
- https://github.com/mratsim/constantine/issues/43

We need to trigger those edge cases more reliably.
5 years ago, on the release of libsecp256k1, the bitcoin developers [described a technique they use](https://www.reddit.com/r/Bitcoin/comments/2rrxq7/on_why_010s_release_notes_say_we_have_reason_to/) to test the library that uncovered a carry bug in OpenSSL [CVE-2014-3570](https://www.openssl.org/news/secadv/20150108.txt). The same kind of carry bug was deemed only discoverable by auditing in TweetNaCL: https://gist.github.com/CodesInChaos/8374632 as it happens with probability 2^-60

The way it works is having the RNG produce long strings of 1 or 0 to trigger carries which when using 2^64 words would happen only with 2^-64 * 2^-64 probability (?).

> The incorrectly squared numbers would be expected to be found randomly with probability around one in 2128, and so when one of the reference implementations of ed25519 had a very similar mistake some described it as "a bug that can only be found by auditing, not by randomized tests". But when we found this weren't auditing OpenSSL (the issue was burred deep in optimized code). Our tests used specially formed random numbers that were intended to explore a class of rare corner cases, a technique I'd previously used in the development of the Opus audio codec. Since our 'random' testing in libsecp256k1 was good enough to find one-in-a-{number too big to name} chance bugs which could "only be found by auditing" I'm a least a little bit proud of the work we've been doing there. (Obviously, we also use many other approaches than random testing on our own code.).

> Part of our testing involves an automatic test harness that verifies agreement of our code with other implementations with random test data, including OpenSSL; though to reduce the chance of an uncontrolled disclosure we backed out some of the low level testing after discovering this bug.
> 
[...]

Issue #43, opened 2020-06-08
Fused modular square root on 32-bit - wrong "isSquare" result


mratsim:
Reproduced locally in 32-bit mode (might not be related to 32-bit as the RNG will initialize differently)

```
test_ec_weierstrass_projective_g1 xoshiro512** seed: 1591646170

[Suite] Elliptic curve in Short Weierstrass form y² = x³ + a x + b with projective coordinates (X, Y, Z): Y²Z = X³ + aXZ² + bZ³ i.e. X = xZ, Y = yZ
  [OK] The infinity point is the neutral element w.r.t. to EC addition
  [OK] Adding opposites gives an infinity point
  [OK] EC add is commutative
  [OK] EC add is associative
    /home/beta/Programming/Nim/constantine/tests/test_ec_weierstrass_projective_g1.nim(170, 19): Check failed: bool(r0 == r1)
  [FAILED] EC double and EC add are consistent
  [OK] EC mul [0]P == Inf
  [OK] EC mul [Order]P == Inf
  [OK] EC mul [1]P == P
  [OK] EC mul [2]P == P.double()
  [OK] EC mul is distributive over EC add
  [OK] EC mul constant-time is equivalent to a simple double-and-add algorithm
Error: execution of an external program failed: '/home/beta/Programming/Nim/constantine/build/test_ec_weierstrass_projective_g1 '
```

https://github.com/mratsim/constantine/blob/082edaa9f51f185a1822da7b9534d11b7b066c71/tests/test_ec_weierstrass_projective_g1.nim#L157-L175

mratsim:
Similar to #30 and #43 another square root bug:

```Nim

import
  # Standard library
  std/[unittest, times],
  # Internals
  ../constantine/config/[common, curves],
  ../constantine/arithmetic,
  ../constantine/io/[io_bigints, io_fields],
  ../constantine/elliptic/[ec_weierstrass_affine, ec_weierstrass_projective]

proc trySetFromCoordX_debug*[F](P: var ECP_SWei_Proj[F], x: F): SecretBool =
  ## Try to create a point the elliptic curve
  ## y² = x³ + a x + b     (affine coordinate)
  ##
  ## The `Z` coordinates is set to 1
  ##
  ## return true and update `P` if `x` leads to a valid point
  ## return false otherwise, in that case `P` is undefined.
  ##
  ## Note: Dedicated robust procedures for hashing-to-curve
  ##       will be provided, this is intended for testing purposes.
[...]

Issue #42, opened 2020-06-07
Fused modular square root on 32-bit - wrong "isSquare" result


mratsim:
![image](https://user-images.githubusercontent.com/22738317/83975776-ee7f5e80-a8f5-11ea-9d52-8e411f5b7064.png)
https://github.com/mratsim/constantine/blob/3d1b1fab9825ec8b5ba5da6cc236b420b408a4be/tests/test_ec_weierstrass_projective_g1.nim#L105-L155

https://travis-ci.com/github/mratsim/constantine/jobs/345566834#L1059

seed: 1591548864

mratsim:
The bug was that somehow the RNG was able to create a random point that was not on the curve.

Point config

```
import
  # Internals
  ../constantine/config/[common, curves],
  ../constantine/arithmetic,
  ../constantine/io/[io_bigints, io_ec],
  ../constantine/elliptic/[ec_weierstrass_affine, ec_weierstrass_projective]

proc test() =

  var a: ECP_SWei_Proj[Fp[BLS12_381]]
  var b: ECP_SWei_Proj[Fp[BLS12_381]]
  var c: ECP_SWei_Proj[Fp[BLS12_381]]

  doAssert a.fromHex(
    x = "0x0426d6f73952e652c229561b9718b76593b60cd9e44aa1f83b4531bb9e6b20a8a959735ac483dbd92dfe48b2ce937987",
    y = "0x025366df8ed339c3e3f8f0a826a24c04343f78c0ec865d56ff1d4718206749a5e07dcb9785520d0f85d4ef209493b41b"
  )
  doAssert b.fromHex(
    x = "0x0d544f56919ea56b16eef9066713331ea54f43ee0d8853d5952847a15694a17d5f21eb9f067e39b3701bebc649ba518a",
    y = "0x0b877f4a58dc1ce70ce9c9d56a536b5cfdff5460b0df60849c9155de252ea25f103f98ac3f9789ffb3d813fe31c3097c",
  )
  doAssert c.fromHex(
    x = "0x065a5dd4ca6a8aba7dc41c8747a99632d6de2f0f3e53899dba906013b3cf2e29a64d7706137b9cf540780235e27035fb",
    y = "0x034c4bd6e37c275f2f7237af190c87e445f624c44b60d0d583c906b2756875e739fd3cc545719b4731577724730571e5"
  )

  var tmp1{.noInit.}, tmp2{.noInit.}: ECP_SWei_Proj[Fp[BLS12_381]]

[...]

Issue #30, opened 2020-06-04
Modular Square Root (spurious?) failure for FKM12_447


mratsim:
https://travis-ci.com/github/mratsim/constantine/jobs/343752157

![image](https://user-images.githubusercontent.com/22738317/83795767-3f961500-a6a0-11ea-9a7b-7e952439b732.png)

Random seed for reproduction is 1591293853, as of commit https://github.com/mratsim/constantine/pull/28/commits/7c71e650a543c90003fe5a35bdd480a5dcc4dda7

mratsim:
Minimal reproducing example:

```Nim
import  ../constantine/[arithmetic, primitives],
        ../constantine/io/[io_fields],
        ../constantine/config/[curves, common]

var a: Fp[FKM12_447]
a.fromHex("0x406e5e74ee09c84fa0c59f2db3ac814a4937e2f57ecd3c0af4265e04598d643c5b772a6549a2d9b825445c34b8ba100fe8d912e61cfda43d")

a.square()

doAssert: bool a.isSquare()
```

PR #59, opened 2020-06-20
[WIP] Skewed RNGs that trigger corner cases

mratsim:
This PR addresses #53.
- An uniform RNG has a hard time triggering carry paths, especially with 64-bit word, the probability is 2^-64 * 2^-64 = 2^-128.
- A significant amount of algebra bugs in established libraries like OpenSSL are due to carries. algebra bugs
- Most fuzzing tools like AFL or LibFuzzer are branch coverage based, but constant-time code is branchless so it's likely that they won't focus on the carry path as it's indistinguishable from the normal path from a Control Flow Graph perspective.

- [x] Add a RNG biaised towards high Hamming Weight words
- [x] Add the same RNG as libsecp256k1 that produces long sequences of 1111... and 0000...
  and found [CVE-2014-3570](https://www.openssl.org/news/secadv/20150108.txt) in OpenSSL and would likely have found similar bug in TweetNaCK: https://gist.github.com/CodesInChaos/8374632
- [x] Use them in testing, in particular check if they can trigger the same edge cases as #30, #42 and #43 that are targeted for solving in #58

### DIFF
--- a/helpers/prng_unsafe.nim
+++ b/helpers/prng_unsafe.nim
@@ -10,7 +10,8 @@ import
   ../constantine/arithmetic/bigints,
   ../constantine/primitives,
   ../constantine/config/[common, curves],
-  ../constantine/elliptic/[ec_weierstrass_affine, ec_weierstrass_projective]
+  ../constantine/elliptic/[ec_weierstrass_affine, ec_weierstrass_projective],
+  ../constantine/io/io_bigints
 
 # ############################################################
 #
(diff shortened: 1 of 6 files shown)


mratsim:
New test failures reported in #60 #61. #61 in particular is probably reproducing the #42 #43 sqrt bug on only 2 limbs which is nice confirmation of the power of this approach.

mratsim:
Mac error raised: https://github.com/mratsim/constantine/issues/64

mratsim:
Test failure raised: https://github.com/mratsim/constantine/issues/62

PR #63, opened 2020-06-20
Use GMP and GNU parallel in GIthub Actions

mratsim:


mratsim:
Raised macos-amd64-c 32-bit failure to https://github.com/mratsim/constantine/issues/64

PR #108, opened 2020-11-28
[WIP] Pairings for bw6 761

mratsim:
This implements pairings for BW6-761

Status: the overall helpers and unoptimized implementations of Miller loop and Final Exponentiation have been set however math debugging must be done.

TODO:

Once the naive implementation is debugged:
- Implement cyclotomic squaring for Fp6
- implement the optimized hard part of final exponentiation.
- Find an addition chain

Ideas on the sources of bug (looking back 1 month later):
1. BW6-761 uses 2 basic miller loops, are we conjugating at the end of those basic miller loop
2. Check the line functions and the coefficients a, b, c (https://eprint.iacr.org/2019/077.pdf) maybe the coefficient computed are misordered. 
image


mratsim:
Pairings are still buggy, both the naive and optimized version. Unsure why to be fixed later, there are some changes that I want merged in the main lib

PR #132, opened 2021-01-23
Square Root & Inversion addition chains - 20% perf increase

mratsim:
This implements addition chains for square root & complete the addition chains for inversions
- [x] BLS12-381
- [x] BLS12-377
- [x] BN254-Nogami
- [x] BN254-Snarks
- [x] BW6-671

On BLS12-381 (of the form `p ≡ 3 (mod 4)`), this provides a 20% speed boost over constant-time exponentiation.
The speed boost depends on the sparsity of the exponent.

## Benefits

1. Square-root is the biggest bottleneck for hashing to curve and the reason why the Hubble Project (https://github.com/thehubbleproject/hubble-contracts) choose to use G1 instead of G2 for signatures (because sqrt on Fp2 requires multiple sqrt computation on G1).
2. It's also a necessary step to deserialize a compressed PublicKey or Signature using `y² = x³ + b` equation. This is a performance bottleneck in Nimbus despite BLST using an addition chain already: https://github.com/status-im/nimbus-eth2/pull/2259
3. The test suite needs to SQRT at least once to create a random point on a curve for fuzzing. Given that 90% of the time is spent testing curve operation, this would speed up the test suite by over 15%. Also the probability of having a point on a curve is about 50% (Euler's criterion: there are (p-1)/2 squares in a field with modulus `p` so we have a probability of ~0.5 to get a good point), so in practice we do an average of 1 + 1/0.5 = 3 squarings. (using infinite power series for mathematicians or the perpetual bond pricing formula for finance/quant).

## Performance

The implementation is about 2% more efficient than BLST for BLS12-81.
- BLST uses 458 operations: https://github.com/supranational/blst/blob/7f6aa706670aa090f0cb69850bc1c388b459c438/src/sqrt-addchain.h
- Constantine uses 449 operations: https://github.com/mratsim/constantine/blob/a4b5f1d1f5d62cb28b31955271512bf002e904b7/constantine/curves/bls12_381_sqrt.nim#L221-L223

On Fp:

On Fp2 before:

On Fp2 after:


mratsim:
BN254 snarks perf also improve by 20%
image

mratsim:
For BLS12-377 which is not in the form `p ≡ 3 (mod 4)`, the improvement is only 10%

image

Issue #131, opened 2021-01-22
Fuzzing failure: Random Inversion testing


mratsim:
https://github.com/mratsim/constantine/runs/1745421711

```
test_finite_fields_powinv xoshiro512** seed: 1611271126

[Suite] Modular exponentiation over finite fields [64-bit mode]
  [OK] n² mod 101
  [OK] x^(p-2) mod p (modular inversion if p prime)

[Suite] Modular division by 2
  [OK] Random modular div2 testing on P224
  [OK] Random modular div2 testing on BN254_Nogami
  [OK] Random modular div2 testing on BN254_Snarks
  [OK] Random modular div2 testing on Curve25519
  [OK] Random modular div2 testing on P256
  [OK] Random modular div2 testing on Secp256k1
  [OK] Random modular div2 testing on BLS12_377
  [OK] Random modular div2 testing on BLS12_381

[Suite] Modular inversion over prime fields [64-bit mode]
  [OK] Specific tests on Fp[BLS12_381]
  [OK] Specific tests on Fp[BN254_Snarks]
    /Users/runner/work/constantine/constantine/constantine/tests/t_finite_fields_powinv.nim(276, 22): Check failed: bool r.isOne()
    /Users/runner/work/constantine/constantine/constantine/tests/t_finite_fields_powinv.nim(278, 22): Check failed: bool r.isOne()
  [FAILED] Random inversion testing on P224
  [OK] Random inversion testing on BN254_Nogami
  [OK] Random inversion testing on BN254_Snarks
  [OK] Random inversion testing on Curve25519
  [OK] Random inversion testing on P256
  [OK] Random inversion testing on Secp256k1
  [OK] Random inversion testing on BLS12_377
  [OK] Random inversion testing on BLS12_381

[Suite] Bug highlighted by property-based testing [64-bit mode]
  [OK] #42 - a^(p-3)/4 (inverse square root)
  [OK] #43 - a^(p-3)/4 (inverse square root)
```

mratsim:
32-bit https://github.com/mratsim/constantine/pull/135/checks?check_run_id=1754998850#step:19:217

```
 test_finite_fields_powinv xoshiro512** seed: 1611433271

[Suite] Modular exponentiation over finite fields [32-bit mode]
  [OK] n² mod 101
  [OK] x^(p-2) mod p (modular inversion if p prime)

[Suite] Modular division by 2
  [OK] Random modular div2 testing on P224
[...]

PR #135, opened 2021-01-23
Fix 50% perf regression (2x with GCC) on binary GCD based inversion

### LINKED ISSUE #134: Perf regression: inversion via extended binary GCD is over 2x slower
In PR #91  we had the following benchmark:

image

In PR #132 we now have

image

Note: I've reduced my CPU max frequency in-between so the exact cycle count cannot be compared, only the relative speed between inversion implementations.

mratsim:
Revert part of #95, fix #134

Before GCC compiler
image

Before Clang
image

After GCC
image

After Clang
image

Conclusion
Implement stuff in assembly ...

### DIFF
--- a/constantine/arithmetic/limbs.nim
+++ b/constantine/arithmetic/limbs.nim
@@ -246,16 +246,22 @@ func cadd*(a: var Limbs, b: Limbs, ctl: SecretBool): Carry =
   ## The carry is always computed whether ctl is true or false
   ##
   ## Time and memory accesses are the same whether a copy occurs or not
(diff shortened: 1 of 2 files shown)


mratsim:
Due to BW6-761 requiring 12 limbs and multiplication cost growing quadratically, the addition chain is 2x more costly than Euclid inversion which grows linearly.

image

So no addition chain for BW6-761.

mratsim:
Failures are fuzzing related and tracked in #131 and #136

PR #155, opened 2021-02-08
Double-Precision towering

mratsim:
This PR reworks the towering to introduce double-precision finite fields.

This moderately accelerate G2  operations and significantly improves Fp12 and pairings computations.

- [x] Double-precision towering: now uses lazy reduction (related to https://github.com/mratsim/constantine/issues/15 and continues https://github.com/mratsim/constantine/pull/144)
- [x] Consistent naming `diffUnr` for unrediced Fp operations instead of `diffNoReduce`
- [x] double-width -> double-precision
- [x]  Consistent naming `diff2xMod` and `diff2xUnr` for double-precision operations.
  - [x] `prod2x`, `square2x` and `redc2x` for double-precision modular arithmetic 
- [x] Removal of `canUseNoCarryMontyMul` and `canUseNoCarryMontySquare` to instead have `getSpareBits` to access the number of extra spare bits in the field element representation.
- [x] Slightly improved assembly for add/sub via interleaving copies and operations and dependency chain latency hiding.
- [x] Small addition chains for Bigint, Fp, FpDbl, extension fields and double-precision extension fields use less temporaries and copies
- [x] Alternate formulae for Fp4[BN254_Snarks] squaring which is 25%, fix #154

PR #158, opened 2021-02-10
Double-precision cubic towering + pairing

mratsim:
This PR continues the work of double precision towering
- https://github.com/mratsim/constantine/pull/72
- https://github.com/mratsim/constantine/pull/144
- https://github.com/mratsim/constantine/pull/148
- https://github.com/mratsim/constantine/pull/155

Hopefully we get so efficient that even BLS24 curves become interesting.

The first commit improves Fp12 multiplication by 5%, which tranlates to about 4% speedup on pairings.

## Fp12 mul, before

image

## Fp12 mul after

image

### DIFF
--- a/constantine/pairing/mul_fp12_by_lines.nim
+++ b/constantine/pairing/mul_fp12_by_lines.nim
@@ -165,7 +165,7 @@ func mul_sparse_by_line_xyz000*[C: static Curve](
     V1.mul2x_sparse_by_x0(f.c1, l.z)
 
     # r1 = (a0 + a1) * (b0 + b1) - v0 - v1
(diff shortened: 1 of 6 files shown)


mratsim:
PR is finished, we obtain a 3% boost in pairings overall.
- The Miller Loop requires a specialized double-precision sparse codepath
- The final exponentiation requires optimized Fp12.
- Some lazy additions were not used but could improve perf (by hundreds of cycles on Fp12). They seem to work for BLS curves, but not for BN curves, is it because of the M-Twist/D-Twist or the fact that there is at least 3 spare bits and so we can store 3p on Fp or 6p on FpDbl, who knows.

## Before

image

## After

image

PR #163, opened 2021-08-13
hash-to-curve BLS12-381 perf

mratsim:
- [x] Fp square no-assembly: split from non-4 non-6 limbs fallback (40% speedup)
- [x] Fast clear cofactor using endomorphism acceleration (5x speedup)
- [x] Fuse mapping 2 points to curve and then adding them
- [x] small optims: noinit, sparse mul
- [x] dedicated invsqrt + cleanup of the squareroot file
- [x] reduce array copies overhead in hash to field


mratsim:
Before a673c91 when compiling without assembly.

image

After

image

mratsim:
Before 0d1e8bb (no assembly i5 Broadwell from 2015)

image

After

image

PR #175, opened 2022-02-13
Low-level refactoring

mratsim:
This refactors the low-level Montgomery rperesentation:
- Fix the ADX autodetection #174, this brings over 30% perf boost
  - Before 
image
  - After: 
image
- Use the convention "operation_domain" like addmod and mulmont.
- Add a specific `fromMont` conversion with x86 assembly and MULX/ADCX/ADOX to speedup by 2x conversion to/from the Montgomery domain

Performance currently sits at:
- 475µs for pairings, 157µs for hashToG2 (Both are needed for BLS verification)
- 60µs for scalar mul on G1 (BLS: Pubkey on G2 and Signature on G1)
- 107µs for scalar mul on G2  (BLS: Pubkey on G1 and Signature on G2) like Ethereum

image

Unfortunately 2 low-level optimizations failed and were stashed.
1. Trying to prefetch data from inputs in Montgomery multiplication by rolling working set of registers and preloading them with the next inputs a couple iterations in advance. Unsure if the memory load didn't have enough execution port next to a MULX/ADCX/ADOX cycle or an issue with filling the instruction cache.
2. Trying to skip the final substraction in addition chains to accelerate square root which is a bottleneck in hashToG2 and point deserialization.


mratsim:
BLST, with Clang and --march=native to ensure ADX on the same machine:

image

image

PR #183, opened 2022-02-26
BLS signatures for Ethereum (BLS sig on BLS12-381 G2 with SHA256)

mratsim:
This adds the first-part of production-grade BLS signatures following https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04.

For now
- [x] serialization,
- [x] deserialization,
- [x] signing,
- [x] verification.

For later:
- [ ] Faster sha256 via SIMD
- [ ] Public key and signature aggregation
- [ ] fastAggregateVerify
- [ ] AggregateVerify
- [ ] Parallelized batch verification

### Performance:

`nimble bench_blssig_on_bls12_381_g2_clang`

image

**BLST** with march-native (for ADX and SHA-NI), with benchmark script https://github.com/status-im/nim-blscurve/blob/0237e4e/benchmarks/bls_signature.nim
`nim c --cc:clang --passC:-march=native --threads:on -d:danger --outdir:build -r --verbosity:0 --hints:off --warnings:off benchmarks/bls_signature.nim`

image

Speedups over BLST:
- 1.14x for signing
- 1.18x for verification

We pass the Ethereum test suite.

Worth noting that CPU features are autodetected which ease packaging for applications.


zah:
It would be valuable to post a longer explanation regarding how this code ended up faster than BLST.

PR #187, opened 2022-03-12
Nimbilize

lantos-lgtm:
Making some of the functions public and starting to make the folder structure confirm with nimble package init structure.
Added jsony, stew and gmp requirements. Note they are not bound to a version atm. 

should `pkg` be removed from all files? example:
-- pkg/jsony
++ jsony 

tests seem to pass but there is this warning which is assumed to fail?
[WARNING] Skipping - deserialization_fails_too_many_bytes.json


lantos-lgtm:
@mratsim

mratsim:
This PR is significantly outdated, would probably be easier to restart from scratch so closing.

PR #195, opened 2022-07-16
Bindings autogeneration via Genny

mratsim:
This is a tentative auto-bindings generation using [Genny](https://github.com/treeform/genny) for:
- C
- Nim
- Javascript
- Python

which would fix #98 (library build) and help #54 (fuzzing)

Currently this is blocked by upstream:
- https://github.com/treeform/genny/issues/43: Add support for generics
- https://github.com/treeform/genny/issues/44: Add support for type aliases
- https://github.com/treeform/genny/issues/45: Add support for generic fields
- https://github.com/treeform/genny/issues/46: Add support for distinct base types

PR #196, opened 2022-07-31
Self-contained bindings generation

mratsim:
This is a first draft at self-contained bindings generation.
Same goal as #195 but without relying on Genny due to blocker issues.

cc @Milerius

The automated header generated from `nim --header` is not clean enough for end-user consumption, so a macro to generate one might be a significant improvement.


bindings/generated/constantine_bls12_381.h
```diff
@@ -0,0 +1,80 @@
+/* Generated by Nim Compiler v1.4.6 */
+/*   (c) 2017 Andreas Rumpf */
+/* The generated code is subject to the original license. */
+#ifndef __constantine_bls12_381__
+#define __constantine_bls12_381__
+#define NIM_INTBITS 64
+
+/* section: NIM_merge_HEADERS */
+
+#include "nimbase.h"
```
mratsim, line 10: Unneeded

bindings/generated/constantine_bls12_381.h
```diff
@@ -0,0 +1,80 @@
...
+#undef near
+#undef far
+#undef powerpc
+#undef unix
+
+/* section: NIM_merge_FRAME_DEFINES */
+
+/* section: NIM_merge_FORWARD_TYPES */
+typedef struct bls12381_fr bls12381_fr;
+typedef struct bls12381_fp bls12381_fp;
```
mratsim, line 29: For proper size in stack the internal fields should be defined. It seems like aliasing breaks that in `type Fr[BLS12_381] = bls12381_fr`

Alternatively can we use exportcpp to properly handle generic exports?

bindings/generated/constantine_bls12_381.h
```diff
@@ -0,0 +1,80 @@
...
+
+/* section: NIM_merge_SEQ_TYPES */
+
+/* section: NIM_merge_FIELD_INFO */
+
+/* section: NIM_merge_TYPE_INFO */
+
+/* section: NIM_merge_PROC_HEADERS */
+N_LIB_PRIVATE N_NOCONV(void, signalHandler)(int sign);
+N_LIB_IMPORT N_CDECL(void, ctt_bls12381_fr_unmarshalBE)(bls12381_fr* dstX60gensym2_, NU8* srcX60gensym2_, NI srcX60gensym2_Len_0);
```
mratsim, line 41: The parameters name shouldn't be mangled, the N_LIB_IMPORT and N_CDECL hurts readability

mratsim: Also NU8* should be `char *` or `uint8_t*` and NI should be `ptrdiff_t`

bindings/generated/constantine_bls12_381.h
```diff
@@ -0,0 +1,80 @@
...
+N_LIB_IMPORT N_CDECL(void, ctt_bls12381_fr_marshalBE)(NU8* dstX60gensym2_, NI dstX60gensym2_Len_0, bls12381_fr* srcX60gensym2_);
+N_LIB_IMPORT N_CDECL(NU64, ctt_bls12381_fr_is_eq)(bls12381_fr* aX60gensym2_, bls12381_fr* bX60gensym2_);
```
mratsim, line 43: SecretBool is NU64 on 64-bit but should be `size_t` in C

bindings/generated/constantine_bls12_381.h
Milerius: > For proper size in stack the internal fields should be defined. It seems like aliasing breaks that in `type Fr[BLS12_381] = bls12381_fr`
> 
> Alternatively can we use exportcpp to properly handle generic exports?

I do think that using export c will simplify the porting of the library in different programming languages. Even if the `exportcpp` is nicer.

PR #207, opened 2022-10-29
Batch additions

mratsim:
This PR follows discussion at Devcon VI with @asn-d6 regarding accelerating validator pubkeys aggregation for large-scale Secret Single Leader Election.

It implements:
- [x] [feat] Batched additions, about 1.8x faster than naive additions starting from 1000+ validators
- [x] [bug] Jacobian mixed addition now handles point doubling
- [x] [bug] Jacobian addition had an aliasing issue if rhs was the infinity point
- [x] [ci] Run tests with -d:release optimizations

Perf 1M points in 180ms on my i9-11980HK (top laptop CPU from 2021)
image

BLST: https://github.com/mratsim/nim-blscurve/pull/1

image

Constantine has 33% speedup over BLST for 1000 point additions (then linear increase)

### DIFF
--- a/benchmarks/bench_ec_g1_batch.nim
+++ b/benchmarks/bench_ec_g1_batch.nim
@@ -48,10 +48,17 @@ proc main() =
     for numPoints in [10, 100, 1000, 10000, 100000, 1000000]:
       let batchIters = max(1, Iters div numPoints)
       multiAddBench(ECP_ShortW_Prj[Fp[curve], G1], numPoints, useBatching = false, batchIters)
+    separator()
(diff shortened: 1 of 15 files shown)


mratsim:
> Constantine has 33% speedup over BLST for 1000 point additions (then linear increase)

Actually there was an issue in the benchmark. I initialized all points to the generator, but that part of the code is not constant-time and affine doubling is significantly slower than affine addition as it involves a squaring: https://github.com/mratsim/constantine/blob/082cd1deb973618c38930e54f7f7b58f3e8835c7/constantine/math/elliptic/ec_shortweierstrass_batch_ops.nim#L203-L215

PR #216, opened 2023-02-07
Move staticFor to the inner repo

mratsim:
Nimble unfortunately removes all folders including tests and benchmarks (and bindings?) on installation. Which is absolutely wrong, especially for libraries tuned to the hardware (bigEndian?).

- https://github.com/nim-lang/nimble/issues/472
- https://github.com/nim-lang/nimble/issues/653
- https://github.com/nim-lang/Nim/issues/6700

In our case, the libraries doesn't work at all because `staticFor` is in the `helpers/` directory. Move it out.

cc @treeform, cc @guzba

_Even nimble needed to use a hack around their recommended `src` structure_ 

_https://github.com/nim-lang/nimble/blob/b629048249ae63117ff90a813037dccab2965983/nimble.nimble#L4-L9_


treeform:
thanks!

PR #222, opened 2023-02-23
Parallel for

mratsim:
This PR adds a `parallelFor` domain-specific language with the following features:
- stride support
- nested parallel loop support
- generic parallel reduction support (i.e., it's possible to use Elliptic Curve addition)
  - Reduction synchronization does not need atomics
  - Reduction result is an awaitable future/FlowVar
- Load balancing is provided through a variation of
  - Lazy binary-splitting: a run-time adaptive work-stealing scheduler.\
    Tzannes, A., G. C. Caragea, R. Barua, and U. Vishkin.\
    In PPoPP ’10, Bangalore, India, January 2010. ACM, pp. 179–190.\
    https://user.eng.umd.edu/~barua/ppopp164.pdf
  - Embracing Explicit Communication in Work-Stealing Runtime Systems.\
    Andreas Prell, 2016\
    https://epub.uni-bayreuth.de/id/eprint/2990/
  We don't just split loops in half, we split based idle workers (unlike OpenMP guided which splits loops depending on workers, idle or busy). TCBV10 mentions that this is expensive due to memory fence, but precise splitting avoids starvation if inner loop is expensive, which is the case in cryptography.
  Instead, we check less often, using log-log iterated backoff, an optimal probabilistic technique for contention/collision resolution in WiFi and Ethernet networks:
  - Adversarial Contention Resolution for Simple Channels\
    Bender, Farach-Colton, He, Kuszmaul, Leiserson, 2005\
    https://people.csail.mit.edu/bradley/papers/BenderFaHe05.pdf
- Memory overhead is:
  - one alloc per loop for parallel for, +1 per split
  - one alloc per loop for parallel reductions, +2 per split
- an experimental feature that will be removed in a following PR: the reserve threads.
  When a worker awaits a future:
  - for latency we want the worker to continue as soon as the future is completed. This might also create more work and expose more parallelism opportunities (in recursive divide-and-conquer algorithm for example)
  - for throughput, and because a scheduler is optimal only when greedy (i.e. within 2x of the best schedule, see Cilk paper), we want an idle thread to take any available work ASAP.
    - but what if that worker ends up with work that blocks it for a long-time? It may lead to work starvation.
  - The experimental feature aimed to solve that by introducing reserve workers. An awaiting worker would give up its active slot and sleep. A reserve worker would wake up and take its place.
  - In theory, the awaiting worker could then be woken up immediately, minimizing latency, and the reserve worker would ensure that we keep using all cores available, maintaining throughput and scheduler greediness.
  - In practice, some benchmarks suffered, the `heat` benchmark had a significant perf regression (7x!!!), fibonacci and DFS (2.5x) bench are slower:
    - before: 
image 
image 
image
   - after: 
image 
image 
image
  - In summary, we have excellent fine-grained parallelism support, but for coarse-grained tasks, awaiting threads are an open problem _if we want to put them to sleep_.


mratsim:
Also huge regression on parallel batch addition / sum reduction:
Before: image
After: image

PR #223, opened 2023-02-24
[Threadpool] Remove reserve threads

mratsim:
This removes a significant threadpool perf pessimization introduced in #222

- reserve threads are removed.
- some perf was also lost because:
  - `primitives.nim` was imported and it needed globals for CPU features detection. Unsure how this exactly slowed a benchmark down though.
  - creating a task is inlined for spawn (but not for loops as those requires more initialization)
  - the conditional branch in `sync` was a debugging aid and is not needed

An additional slight optimization is avoiding syscalls in `wake` when we have sleepy workers that are not yet asleep. The change in epoch is sufficient to avoid their sleep.

Finally, the stealHalf and adaptative theft policy code has been removed as it was never finished.

### DIFF
--- a/constantine.nimble
+++ b/constantine.nimble
@@ -243,6 +243,8 @@ const testDescNvidia: seq[string] = @[
 const testDescThreadpool: seq[string] = @[
   "constantine/platforms/threadpool/examples/e01_simple_tasks.nim",
   "constantine/platforms/threadpool/examples/e02_parallel_pi.nim",
+  "constantine/platforms/threadpool/examples/e03_parallel_for.nim",
+  "constantine/platforms/threadpool/examples/e04_parallel_reduce.nim",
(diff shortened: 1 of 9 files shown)


mratsim:
# Benchmarks

While many are unaffected, some are showing dramatic improvements

## Log-sum-exp (up to **58x** speedup)

Yup 58x speedup

Before 
image

after 
image

## Heat diffusion (7.11x faster)

Before 
image


After 
image

## Depth-First Search (up to 2x speedup)

Before 
image


After 
image

## Fibonacci (1.45x faster)

Before 
image


After 
image

# Crypto bench (6x improvement)

Focus on batch elliptic addition / sum reduction of 65536 BLS12-381 points

Before image

After 
image

The pessimization code was slower than serial while after fix the code is 5.9x faster than serial

PR #224, opened 2023-02-25
[Threadpool] Backoff revamp

mratsim:
This reworks the backoff of the Threadpool

1. The global backoff didn't actually put thread to sleep, just blocked them in a spin loop, but the futex value wasn't set properly.
2. There was a hard to trigger race condition when awaiting a task that wasn't completed: https://github.com/mratsim/constantine/blob/1dfbb8bd4f2fefc4059f4aefb040f0b3ba2918cb/constantine/platforms/threadpool/threadpool.nim#L260-L269
    1. The very last part of running the task with `task.fn(closureEnv)` is setting its status to complete.
    2. Then the task runner checks if the task has a future, and if yes if a waiter was registered and wake it.
    3. But as soon as a task is complete, a waiter can copy its result, and once done free the task.
    4. If the task is freed before the task runner finishes the future and waiter check, we get a use-after-free.

Issue 1 is solved by using the futex itself for epoch tracking / counting events in the Eventcount. At the same time, the Eventcount now always supports up to 65536 threads instead of 256 (32-bit) and 65536 (64-bit)

Issue 2 is solved by adding an additional `canBeFreed` flag to the task and "separate" setting the task to complete and telling the waiter that it can be freed: https://github.com/mratsim/constantine/blob/1a914abf2bef5924536e7f2140a7743e0ab44cd7/constantine/platforms/threadpool/threadpool.nim#L267-L271. Additionally we completely removed the EventNotifier / local backoff and directly use the task as a backoff data structure. While before we needed:
- In the task: `completed: Atomic[bool]`, `thiefID: Atomic[int32]`,  `waiter: Atomic[ptr EventNotifier]`
- In the EventNotifier: 3 bools and 1 uint32
- Now we only need 2x uint32 to track thief, waiter, canBeFreed and task completion
- The new  task state can supports up to 32768 threads.


mratsim:
# Benchmarks

## 1. LogSumExp (Parallel For loop reduction), 25% to 5x speedup
ksnip_20230225-163611
ksnip_20230225-163333


## 2. Heat (Recursive divide-and-conquer), 33% speedup and no use-after-free
ksnip_20230225-163904
ksnip_20230225-163748

## 3. Fibonacci (Scheduler overhead, with empty tasks), 25% slowdown
ksnip_20230225-164009
ksnip_20230225-163856

## 3. Depth-First Search (Scheduler overhead, with somewhat empty tasks), 15% slowdown
ksnip_20230225-164105
ksnip_20230225-165401

## 4. Single Task Producer: 1M tasks, 10 us each, 6% speedup

ksnip_20230225-164246
ksnip_20230225-164545

## 5. Bouncing Producer-Consumer: 10k depth, 999 tasks per depth, 1us per task (no change)
ksnip_20230225-164404
ksnip_20230225-164459

## Analysis

Except for the almost empty tasks (fibonacci and DFS), we see improvement as soon as a task takes 1us and very significant improvement for long tasks. The runtime is now as fast or significantly faster than Weave (Log-Sum-Exp bench) on parallel loop.
For tasks:
- it's as fast as Weave for easy to schedule tasks (heat / parallel divide-and-conquer)
- it suffers if producers-consumers have severe imbalance, Weave is 8x faster on the single task producer case with 1M tasks 10us each. And probably we can construct severe scenarios with bouncing-producer-consumer. The reason why is very likely the lack of `stealHalf` operation that can logarithmically redistribute all tasks in a runtime.

Thankfully we should be good enough for our use cases: batch addition / sum reduction (a loop) and Multi-Scalar Multiplication (parallel loops)

Issue #230, opened 2023-04-18
GCC miscompiles Fp6 Frobenius with -flto flag


mratsim:
After applying fixes for #229 in #228 548cf20

```
nim c --cc:gcc -r -d:release --passC:-flto=auto --passL:-flto=auto --outdir:build tests/math/t_fp6_frobenius.nim
```

fails with 
```md
test_fp6_frobenius xoshiro512** seed: 1681811972

[Suite] 𝔽p6 Frobenius map: Frobenius(a, k) = a^(pᵏ) (mod p⁶) [64-bit words]
    Frobenius(a) for Fp6[BN254_Nogami]
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    Frobenius(a) for Fp6[BN254_Snarks]
    Frobenius(a) for Fp6[BLS12_377]
    Frobenius(a) for Fp6[BLS12_381]
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    /[...]/constantine/tests/math/t_fp_tower_frobenius_template.nim(80, 21): Check failed: bool(a == fa)
    Frobenius(a) for Fp6[BW6_761]
[...]

Issue #229, opened 2023-04-17
GCC miscompiles Fp2 with --opt:size flag


mratsim:
From #228,

This fails
```
nim c --cc:gcc -r -d:release --opt:size --outdir:build tests/math/t_fp2.nim
```
Those do not
```
nim c --cc:clang -r -d:release --opt:size --outdir:build tests/math/t_fp2.nim
nim c --cc:gcc -r -d:release --outdir:build tests/math/t_fp2.nim
```

The function that is miscompiled is https://github.com/mratsim/constantine/blob/93dac2503c2b36ed667e03dabc479933b33bc6df/constantine/math/extension_fields/assembly/fp2_asm_x86_adx_bmi2.nim#L117-L135

mratsim:
Following further investigation in #230, I believe that the compiler finds out that M is a constant by constant folding through compilation units with LTO and generating new procedures with only the variable input
As the inline assembly directly use the pointer as an input register, it's likely that its value has been hardcoded. This wouldn't be a problem if the address of the constant didn't change but who knows what the linker does (and there is relocatable code and Address Space Layout Randomization as well)

PR #231, opened 2023-04-23
Rework assembly to be compatible with LTO

mratsim:
This PR fixes the core problems behind #229 and #230 (instead of just tagging {.noInline.}
)

## Write up

Inline assembly has several constraints to guide the compiler in register allocation, in particular `r` (register) and `m` (memory).

The `memory` operand has an issue, as tested in code that predated the assembler, for an offset of 8 bytes, GCC supports the syntax `8+%[M]` while Clang supports `8%[M]`. `%` means a symbol that will be replaced and `[M]` might be replaced by a dereferenced pointer `(rax)` leading to `8(rax)` with Clang, the proper AT&T assembly syntax for base pointer + displacement or `8+(rax)` in GCC, which works as a specific workaround.

Due to this, Constantine use to use the register constraint instead storing pointers, and use the memory operand only has a workaround for running out of registers, and only for a ingle uint64.

But with LTO, compilers can do constant propagation, and hardcode the `register` value, not realizing that it is a pointer to memory, which then points to invalid memory once the linker reorgs the binary after LTO.

1. An easy workaround is to tag the affected functions as noinline.
2. Another workaround is to change the register constraint to memory generating different code for GCC and Clang.

Solution 1 would defeat the purpose of LTO in a hot function.

Solution 2 unfortunately doesn't work. With constant propagation made visible thanks to LTOm  the compiler might replace `%[M]` by `MyModulus(%rip)` to indicate to the linker to paste the relative address of a constant. This compose poorly with `8%[M]` syntax with Clang leading to undeclared identifier. A clever solution might be to have each cell as an input with its own identifier like `%[M0]`, `%[M8]` and so on. But then compilers complain with asm has impossible constraints for GCC, and recoloring failure or not enough registers for Clang.

The last solution adopted here is to use Intel syntax for the assembler as it plays well with inline assembly copy-paste `QWORD ptr [%M + 8]`.

This require Clang 14. And Apple Clang does not support it (bug raised to Apple FB12137688).


mratsim:
TODO: 3% perf regression on signatures on BLS12-381 G2 (see the deleted perf section in the README). Some `inline` functions were replaced by non-inline and left for the compiler+LTO to optimize, the compiler probably didn't inline those. Besides `inline`, it's possible that `attribute((hot))`` might be an interesting alternative property.

mratsim:
The cyclotomic test failing on windows (`nim c -r  --passC:-Wno-stringop-overflow --passL:-Wno-stringop-overflow  --passC:-flto=auto --passL:-flto=auto  -d:danger  --panics:on -d:noSignalHandler  --mm:arc -d:useMalloc  --verbosity:0 --hints:off --warnings:off  --passC:-fno-semantic-interposition  --passC:-falign-functions=64  --outdir:build/testsuite  --nimcache:nimcache/tests/math/t_pairing_cyclotomic_subgroup.nim tests/math/t_pairing_cyclotomic_subgroup.nim`) cannot be reproduced with MinGW64 + GCC 8.1 but can with MinGW64 + GCC 12.2

PR #239, opened 2023-05-24
Ethereum KZG polynomial commitments / EIP-4844

mratsim:
Writeup upcoming

Spec: https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/deneb/polynomial-commitments.md
Tests: https://github.com/ethereum/consensus-spec-tests/tree/v1.3.0/tests/general/deneb/kzg
Ref impl: https://github.com/ethereum/c-kzg-4844


mratsim:
Failures due to an upstream regression: https://github.com/nim-lang/Nim/issues/21958

mratsim:
The EIP 4844 / Ethereum KZG PR has become too huge and will be split in at least 2 parts.
We will merge this part, with minimal testing, as it will serve as an example as Constantine public API for serialization of Banderwagon and the library structure for IPA commitments (cc @zah)

The first part introduces:
- a significant refactoring of the serialization, codecs primitives and their error models. This impact the Ethereum BLS Signatures API.
  - A per-curve codec file, for example `codecs_bls12_381` https://github.com/mratsim/constantine/pull/239/files#diff-c65f7dc4a0b9580364db8ac21749f5a065ffcfa4d1441543e6cbfd5b8f1caf58 is created to handle serialization.
    - Unfortunately serialization cannot be generic as BLS12-381 serialization uses extra 3 bits (384-bit available for 48 bytes) to indicate: compression, sign, infinity. This cannot be replicated on BN254 (only 2 bits) or even Banderwagon (only 1 bit available)
  - The protocol is supposed to rewrap the curve codec error into a protocol error. Feedback welcome.
  - C wrappers now warn on unused status codes
- The internal primitives (working on deserialized types) have all be implemented and significantly optimized, in particular:
  -  bit-reversal permutations is 2x~3x faster than reference c-kzg-4844 by optimizing cache line usage. _Note: this can be pre-computed and cached for Ethereum, but might be critical for Plonkish arithmetization in ZK provers_
  - KZG proof generation caches inverse generation
- strided views (mono-dimensional tensors) for zero-copy FFTs.
- Polynomials, lagrange interpolation and Barycentric formulas 
image
- a binary trusted setup interchange format specification, tuned for loading speed, including with SIMD, https://github.com/mratsim/constantine/blob/48ecc5cec88696ef46994a87f6f46b70069d2622/constantine/trusted_setups/README.md
- a SAGE generator for trusted setups related constants (primitives roots of unity, ...)

PR #240, opened 2023-05-29
Path reorgs

mratsim:
- split `tests/math`, it has become very big and unwieldly
- move threadpool out of platforms: it can be a standalone package in a monorepo


mratsim:
Failures due to an upstream regression: https://github.com/nim-lang/Nim/issues/21958

mratsim:
Merging despite failures due to https://github.com/nim-lang/Nim/issues/21958

This is a no-risk PR with only changes in imports and file paths for tests and threadpool.

Works on my machine:tm:

PR #242, opened 2023-05-30
Modular exponentiation (arbitrary output) and EIP-198

mratsim:
While Constantine supports modular exponentiation (as used in RSA or Ethereum EIP-198 https://eips.ethereum.org/EIPS/eip-198):
- it only supports it for modulus with size / number of bits known at compile-time
- and only for odd integers

This makes it only useful for cryptography.

Complaints about the poor state of bigints have been occuring every once in a while in Nim, for example:
- https://forum.nim-lang.org/t/7276
- https://github.com/status-im/nimbus-eth1/issues/1584

To solve this, I outlined last week a guide on how to implement fast modular exponentiation at https://github.com/status-im/nim-stint/issues/126.
This PR adds a new arbitrary precision backend to Constantine following this guide, and will use this for EIP-198.

Perf notes:
- the code is generic to any size
- the code is NOT constant-time
- it reaches 80% to 120% of GMP speed **without assembly** on 256-bit integers (for the Ethereum Virtual Machine)
- the code does not use assembly at the moment. I expect ~30% performance gain with Clang (more on GCC due to GCC inefficient carry handling)
- a constant-time modulo reduction is used in the hot path, a variable-time alternative that uses hardware division might be significant (or not)
- sliding window exponentiation might gain an extra 15% to 20%


mratsim:
API & organization note:

Currently 2 APIs for arbitrary precision views are cohabiting:
- ptr UncheckedArray + number of bits
- openArray

The first one is inherited from bigint division. The first one was introduced in this PR. We likely want to pick one in the future.

## Benchmarks

A quick-and-dirty benchmark to get a sense of the perf vs GMP ranked the new `powmod` as fast or up to 40% faster than GMP on 256-bit inputs: https://github.com/status-im/nim-stint/issues/126#issuecomment-1567181309

bench 1

A more thorough benchmarks gives the avantage to GMP, but Constantine still reaches over 85% of GMP performance (without assembly or advanced algorithm like sliding windows and still with a constant-time modulo in the mix)

image

Also the crypto backend, for fixed-precision known at compile-time and odd moduli is 2x faster.

mratsim:
Writeup on the performance issues in Stint and nim-bigints at https://forum.nim-lang.org/t/7276#67901

PR #249, opened 2023-07-02
Accelerate eth_evm_modexp by 25x by dividing input size by 8

mratsim:
Found by @guidovranken in the Ethereum Foundation sponsored fuzzing campaign,

eth_evm_modexp is very slow on some input.

The issue is located here:

1. We read the length in bytes here: https://github.com/mratsim/constantine/blob/d0f4ad8cdaf5fbbec98889bfcfb3a04082ff6f94/constantine/ethereum_evm_precompiles.nim#L425-L428
2. But we allocated as many 64-bit words as there were bytes instead of dividing by 8: https://github.com/mratsim/constantine/blob/d0f4ad8cdaf5fbbec98889bfcfb3a04082ff6f94/constantine/ethereum_evm_precompiles.nim#L455-L457

The issue is that multiplication scales quadratically, about 1.5n² with the number of words. In our case the ratio before/after is 25x leading to a DOS vulnerability


mratsim:
Exponentiation with even base and power-of-two modulus is now instant for a lot more moduli.

In particular the DOS vector has been accelerated by 2310x.

Before:
image

After generic fix (i.e. dividing input size by 8) - 28x acceleration.
image

Using special structure - 2370x acceleration.
image

PR #251, opened 2023-07-10
fix overflow when truncating in submod2k

mratsim:
Fix @guidovranken fuzz failure 8

And separate stack hardening and sanitizers in test cases.
Enable stack hardening on all tests and sanitizers on modexp tests due to the frequent uses of `allocStackArray`


constantine/math_arbitrary_precision/arithmetic/limbs_mod2k.nim
```diff
@@ -43,14 +43,17 @@ func submod2k_vartime*(r{.noAlias.}: var openArray[SecretWord], a, b: openArray[
...
+      "  r.len: " & $r.len & "\n" &
+      "  k: " & $k & "\n" &
+      "  k/WordBitWidth: " & $(k.int shr SlotShift) &
+      "\n" # [AssertionDefect]
 
   # We can compute (mod 2ʷ) with w >= k
   # Hence we truncate the substraction to the next multiple of the word size
   template trunc(x: openArray[SecretWord]): openArray[SecretWord] =
-    x.toOpenArray(0, k.int.wordsRequired()-1)
+    let truncHi =  min(x.len, k.int.wordsRequired()) - 1
```
mratsim, line 55: The absence of the `min` was the main bug

Issue #253, opened 2023-08-14
Scalar-mul: windowed NAF + endomorphism acceleration


mratsim:
Currently Constantine implements the following acceleration for scalar multiplication:

1. Constant-time GLV-SAC representation + endomorphism acceleration
2. Variable-time Windowed NAF

For secret data, approach 1 is fast and has a small memory footprint.

However, it is slow for FFTs as used in polynomial commitment schemes to switch from polynomials in monomial basis to polynomials in Lagrange basis.

This happens because lots of small factors (including 1s) appear during FFTs that constant-time multiplication cannot skip. (joint investigation with KZG implementers (benjaminion/protolambda) in Feb 2021 that in particular led to counting bytes in BLST https://github.com/benjaminion/c-kzg/commit/f557f32ed113af4c7e38db6987b8e0751d100f61 to match MCL speed.

So for maximum perf for FFTs, we want wNAF + endomorphism acceleration.

PR #258, opened 2023-08-28
Accelerate FFT

mratsim:
- [x] vartime scalarmul: endomorphism support and short-circuit on unitary scalar => 30% perf improvement

_Before_
image

_After_
image


mratsim:
In the end the PR cuts FFT time by 58%.

image

There is an extra possible optimization to followup on #253
One of the roots of unity is `0x8d51ccce760304d0ec030002760300000001000000000000`
which is 192 bits long. The endomorphism split it in a (127-bit, 62-bit) pair of wNAF-4 encoded scalars. However the compute time is the same as a full 255-bit scalar.

PR #268, opened 2023-09-08
modexp: 2.5x accel on small exponent

mratsim:
Improve a potential DOS vector for the modexp precompile.
The overhead of computing r² was too high, changing to the `reduce` operation compared to repeated doubling (which is used for the compile-time proc because `reduce` requires pointers) improves perf by 2.5x.

Todo:
- Provide a vartime reduction to improve perf further
- Use it at compile-time to improve compilation speed
- Evaluate Barrett Reduction for small modexp instead of Montgomery
- Or potentially just multiplication and modular reduction


mratsim:
Followed up by #286

PR #271, opened 2023-09-12
feat: add banderwagon

### LINKED ISSUE #263: Add Banderwagon
### Banderwagon is a scheme built on top of Bandersnatch to solve the cofactor issue of Twisted Edward curves.
For the details of the implementation approach refer : 
- [Bandersnatch Implementation Notes](https://hackmd.io/wliPP_RMT4emsucVuCqfHA?view)
- [From Bandersnatch to Banderwagon](https://hackmd.io/@6iQDuIePQjyYBqDChYw_jg/BJBNcv9fq)

The Decaf-Style implementation should be used. The implementation would add the following
- Banderwagon Curve ( Same curve equation and constants as Bandersnatch )
- Override the `==` equality check, to the equality check mentioned in the implementation notes
- Add the subgroup check, to check if the point is in the quotient subgroup
- This is to be used to check if the point lies in the Banderwagon while importing a point from serialized bytes, thus serialization support needs to be added

***The combination of checking if $`1-aX^2`$ is a square, and the serialisation strategy over bandersnatch is what is known as banderwagon.***

advaita-saha:
Fixes #263 
Adds all the features mentioned in the issue

Other than that, the following have been changed :
- Status codes have been shifted to a common `codecs_status_codes.nim` file
- `trySetFromCoordX` added to Twisted Edwards curve, both projective & affine


constantine/serialization/codecs_banderwagon.nim
```diff
@@ -0,0 +1,111 @@
...
+  ],
+  ../math/[
+    extension_fields,
+    arithmetic,
+    constants/banderwagon],
+  ../math/io/[io_bigints, io_fields],
+  ./codecs_status_codes
+
+type
+  EC_P* = ECP_TwEdwards_Prj[Fp[Banderwagon]]
```
mratsim, line 30:
```suggestion
  ECP_Prj* = ECP_TwEdwards_Prj[Fp[Banderwagon]]
```

for Elliptic Curve Point Projective

constantine/serialization/codecs_banderwagon.nim
```diff
@@ -0,0 +1,111 @@
...
+  EC_A* = ECP_TwEdwards_Aff[Fp[Banderwagon]]
```
mratsim, line 31:
```suggestion
  EC_Aff* = ECP_TwEdwards_Aff[Fp[Banderwagon]]
```

for elliptic curve point affine

A and P are a bit too short

constantine/serialization/codecs_banderwagon.nim
```diff
@@ -0,0 +1,111 @@
...
+
+func serialize*(dst: var array[32, byte], P: EC_P): CttCodecEccStatus =
+  ## Serialize a Banderwagon point(x, y) in the format
+  ## 
+  ## serialized bits = 256
+  ## 1st bit -> sign(y)
+  ## remaining 255 bits -> x ( big endian format )
+  ## 
+  ## Returns cttCodecEcc_Success if successful
```
mratsim, line 40:
```suggestion
  ## Returns cttCodecEcc_Success if successful
  ## Spec: https://hackmd.io/@6iQDuIePQjyYBqDChYw_jg/BJBNcv9fq#Serialisation
```

constantine/serialization/codecs_banderwagon.nim
```diff
@@ -0,0 +1,111 @@
...
+
+func deserialize_unchecked*(dst: var EC_P, src: array[32, byte]): CttCodecEccStatus =
+  ## Deserialize a Banderwagon point (x, y) in format
+  ## 
+  ## serialized bits = 256
+  ## 1st bit -> sign(y)
+  ## remaining 255 bits -> x ( big endian format )
+  ## 
+  ## Returns cttCodecEcc_Success if successful
+  ##
```
mratsim, line 67:
```suggestion
  ## https://hackmd.io/@6iQDuIePQjyYBqDChYw_jg/BJBNcv9fq#Serialisation
```

constantine/math/constants/banderwagon.nim
```diff
@@ -0,0 +1,69 @@
...
+
+  # Check for the zero points
+  result = not(P.x.is_zero() and P.y.is_zero())
+  result = result or not(Q.x.is_zero() and Q.y.is_zero())
+
+  ## Check for the equality of the points
+  ## X1 * Y2 == X2 * Y1
+  lhs.prod(P.x, Q.y)
+  rhs.prod(Q.x, P.y)
+  result = result and lhs == rhs
```
mratsim, line 35: Interesting, it sems like this formula avoiding using Z coordinates and would be faster than the original one:

https://github.com/mratsim/constantine/blob/d51699248db04e29c7b1ad97e0bafa1499db00b5/constantine/math/elliptic/ec_twistededwards_projective.nim#L35-L47

Might be intersting to change the original and check the tests passes

constantine/math/constants/banderwagon.nim
```diff
@@ -0,0 +1,69 @@
...
+
+func map_to_field*(P: ECP_TwEdwards_Prj[Fp[Banderwagon]]): Fp =
+  ## Map a point on the Banderwagon curve to the base field
+  ## 
+  result.prod(P.x, P.y.inv())
```
mratsim, line 40: This can be put in https://github.com/mratsim/constantine/blob/master/constantine/hash_to_curve/hash_to_curve.nim

There will be specialized per curve mappings

mratsim (commented):
Looks good so far

constantine/math/constants/banderwagon.nim
mratsim: Actually misread, I thought it was to map random bytes to a field.

How is this used? I don't remember seeing a spec/usage for this.

constantine/math/constants/banderwagon.nim
advaita-saha: I think we can try this out separately. And keep that out of this PR

constantine/math/constants/banderwagon.nim
advaita-saha: I kinda had a very similar doubt regarding this, but I saw this in the notes of both gottfried and kev. So thought I should include this.

But I haven't found any usage for this anywhere till now. What do you suggest we should do?

constantine/math/constants/banderwagon.nim
mratsim: The original formulas should still work for Banderwagon unless I missed something though so this shouldn't be needed.

On a side note if someone imports elliptic/ec_twistededwards_projective.nim and does not import constants/banderwagon.nim (or it's not reexported) the wrong `==` will be called.

So the file should import/reexport it to ensure the specialization is called.

constantine/math/constants/banderwagon.nim
mratsim: I think we can remove this at the moment, it's very short and can be reintegrated in a place that makes sense if needed.

constantine/math/constants/banderwagon.nim
```diff
@@ -0,0 +1,69 @@
...
+  one.setOne()
+  res.setZero()
+
+  ## Compute 1 - aX^2 and check its legendre symbol
+  res.prod(P.x, P.x)
+  res.prod(res, Banderwagon.getCoefA())
+  res.neg(res)
+  res.sum(res, one)
+
+  return res.isSquare()
```
mratsim, line 69: This should be in `constants/banderwagon_subgroups.nim`

constantine/math/elliptic/ec_twistededwards_affine.nim
```diff
@@ -147,3 +192,29 @@ func cneg*(P: var ECP_TwEdwards_Aff, ctl: CTBool) =
...
+
+# ############################################################
+#
+#              Banderwagon Specific Operations
+#
+# ############################################################
+
+func `==`*(P, Q: ECP_TwEdwards_Aff[Fp[Banderwagon]]): SecretBool =
+  ## Equality check for points in the Banderwagon Group
+  ## The equals method is different for the quotient group
```
mratsim, line 204:
```suggestion
  ## The equality check is optimized for the quotient group
  ## see: https://hackmd.io/@6iQDuIePQjyYBqDChYw_jg/BJBNcv9fq#Equality-check
```

constantine/math/elliptic/ec_twistededwards_projective.nim
```diff
@@ -308,3 +328,29 @@ func projective*[F](
...
+
+# ############################################################
+#
+#              Banderwagon Specific Operations
+#
+# ############################################################
+
+func `==`*(P, Q: ECP_TwEdwards_Prj[Fp[Banderwagon]]): SecretBool =
+  ## Equality check for points in the Banderwagon Group
+  ## The equals method is different for the quotient group
```
mratsim, line 340:
```suggestion
  ## The equality check is optimized for the quotient group
  ## see: https://hackmd.io/@6iQDuIePQjyYBqDChYw_jg/BJBNcv9fq#Equality-check
```

mratsim (approved):
LGTM, we only need tests now

advaita-saha:
@mratsim While reviewing please have a close look at the precompute for `primeMinus1div2` as I am not very sure how good the implementation I did for that

constantine/math/config/precompute.nim
```diff
@@ -392,6 +392,16 @@ func primePlus1div2*(P: BigInt): BigInt =
...
 
+func primeMinus1div2*(P: BigInt): BigInt =
+  ## Compute (P-1)/2
+  ## For use in constant-time modular inversion
+  ##
+  ## Warning ⚠️: Result is in the canonical domain (not Montgomery)
+
+  result = P
+  discard result.sub(1)
+  result.shiftRight(1)
```
mratsim, line 403: Nitpick: actually let's pick 6 and 7, in binary they are `0b110` and `0b111`, after a right shift they both become `0b11` which is 3, so the -1 is free for odd integers

constantine/math/config/precompute.nim
```diff
@@ -392,6 +392,16 @@ func primePlus1div2*(P: BigInt): BigInt =
...
   doAssert not carry
 
+func primeMinus1div2*(P: BigInt): BigInt =
+  ## Compute (P-1)/2
+  ## For use in constant-time modular inversion
+  ##
+  ## Warning ⚠️: Result is in the canonical domain (not Montgomery)
+
+  result = P
+  discard result.sub(1)
```
mratsim, line 402:
```suggestion
  # discard result.sub(1) # right-shifting automatically implies "-1" for odd numbers (which all prime >2 are).
```

constantine/math/elliptic/ec_twistededwards_affine.nim
```diff
@@ -76,6 +81,51 @@ func isOnCurve*[F](x, y: F): SecretBool =
...
+    else:
+      P.y.fromUint uint -F.C.getCoefD()
+      P.y.neg()
+  else:
+    P.y = F.C.getCoefD()
+  P.y *= t
+  P.y.neg()
+  P.y += one
+
+  # (1 - ax²)
```
mratsim, line 110: a future TODO, a = 5 and we should shortcut when a <= 15 to use the addition chains here: https://github.com/mratsim/constantine/blob/7b64f85a29c466b9f3b827059b9587a7a39d979b/constantine/math/arithmetic/finite_fields.nim#L475-L479

mratsim (approved):
Excellent

PR #278, opened 2023-09-24
`MapToScalarField()` added for Banderwagon points

### LINKED ISSUE #277: Map To Scalar Field
### Mapping a $F_p$ point to  $F_r$  for *Banderwagon* Elements

This is specifically required for the ***Ethereum Verkle Trie***, implementations where this will be needed.
This feature will require us to implement the following
1. A `MapToField()` function that will return the $\frac{x}{y}$ value. This creates an injective map from the banderwagon quotient group to the field $F_p$. Please find the detailed specs [here](https://hackmd.io/@6iQDuIePQjyYBqDChYw_jg/BJBNcv9fq#Map-To-Field)
2. The `MapToScalarField()` function, this will take the $\frac{x}{y}$ value from the above function and convert that to bytes in Little Endian or Big Endian, and then load that to an $F_r$ element

advaita-saha:
Fixes #277 

Refere the issue #277 for the detailed specs

- [x] Write the `mapToBaseField()` function for computing $\frac{x}{y}$ and returning as an $F_p$ element
- [x] `MapToScalarField()` for  converting the $F_p$ element to $F_r$, in between it is done by first converting $F_p$ -> *Bytes* -> $F_r$
- [x] Tests for `mapToBaseField()` & `MapToScalarField()`
- [x] Commenting and linking specs documents in the code 
- [x] Adding the `BatchMapToField()`, this does the same operations but for a batch of *Banderwagon* points together
- [x] Tests for the Batch operations
- [x] Final checks, fixes and commenting the codebase

### DIFF
--- a/constantine/ethereum_verkle_primitives.nim
+++ b/constantine/ethereum_verkle_primitives.nim
@@ -15,7 +15,10 @@
 import
   ./math/config/[type_ff, curves],
   ./math/arithmetic,
-  ./math/elliptic/ec_twistededwards_projective,
+  ./math/elliptic/[
+    ec_twistededwards_projective,
+    ec_twistededwards_batch_ops
+    ],
   ./math/io/[io_bigints, io_fields],
   ./curves_primitives
   
@@ -48,3 +51,25 @@ func mapToScalarField*(res: var Fr[Banderwagon], p: ECP_TwEdwards_Prj[Fp[Banderw
   let check2 = res.unmarshalBE(baseFieldBytes)      # bytes -> Fr
 
   return check1 and check2
+
+func batchMapToScalarField*[N: static int](
+      res: var array[N, Fr[Banderwagon]], 
+      points: array[N, ECP_TwEdwards_Prj[Fp[Banderwagon]]]): bool {.discardable.} =
+
+  var ys: array[N, Fp[Banderwagon]]
+
+  for i in 0 ..< N:
+    ys[i] = points[i].y
+  
+  ys.batchInvert_vartime()
+
+  var check: bool
+  for i in 0 ..< N:
+    var mappedElement: Fp[Banderwagon]
+    var bytes: array[32, byte]
+
+    mappedElement.prod(points[i].x, ys[i])
+    check = bytes.marshalBE(mappedElement)
+    check = check and res[i].unmarshalBE(bytes)
+
+  return check  
--- a/constantine/ethereum_verkle_primitives.nim
+++ b/constantine/ethereum_verkle_primitives.nim
@@ -52,10 +52,24 @@ func mapToScalarField*(res: var Fr[Banderwagon], p: ECP_TwEdwards_Prj[Fp[Banderw
 
   return check1 and check2
 
+## ############################################################
+##
+##                      Batch Operations
+##
+## ############################################################
 func batchMapToScalarField*[N: static int](
       res: var array[N, Fr[Banderwagon]], 
       points: array[N, ECP_TwEdwards_Prj[Fp[Banderwagon]]]): bool {.discardable.} =
-
+  ## This function performs the `mapToScalarField` operation
+  ## on a batch of points
+  ## 
+  ## The batch inversion used in this using
+  ## the montogomenry trick, makes is faster than
+  ## just iterating of over the array of points and 
+  ## converting the curve points to field elements
+  ## 
+  ## Spec : https://hackmd.io/wliPP_RMT4emsucVuCqfHA?view#MapToFieldElement
+  
   var ys: array[N, Fp[Banderwagon]]
 
   for i in 0 ..< N:
--- a/constantine/ethereum_verkle_primitives.nim
+++ b/constantine/ethereum_verkle_primitives.nim
@@ -20,13 +20,14 @@ import
   ./curves_primitives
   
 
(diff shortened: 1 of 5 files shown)


constantine/ethereum_verkle_primitives.nim
```diff
@@ -0,0 +1,35 @@
...
+  ./math/io/[io_bigints, io_fields],
+  ./curves_primitives
+  
+
+func mapToBaseField*(p: ECP_TwEdwards_Prj[Fp[Banderwagon]]): Fp[Banderwagon] =
+  var invY: Fp[Banderwagon]
+  invY.inv(p.y)
+  result.prod(p.x, invY)
+
+func MapToScalarField*(res: var Fr[Banderwagon], p: ECP_TwEdwards_Prj[Fp[Banderwagon]]): bool {.discardable.} =
```
mratsim, line 28: The name should start with a lowercase `m`

constantine/ethereum_verkle_primitives.nim
```diff
@@ -0,0 +1,35 @@
...
+
+import
+  ./math/config/[type_ff, curves],
+  ./math/arithmetic,
+  ./math/elliptic/ec_twistededwards_projective,
+  ./math/io/[io_bigints, io_fields],
+  ./curves_primitives
+  
+
+func mapToBaseField*(p: ECP_TwEdwards_Prj[Fp[Banderwagon]]): Fp[Banderwagon] =
```
mratsim, line 23: don't return a result, this is likely to cause slowness without NRVO (Named Return Value Optimization) / copy-elision: https://github.com/nim-lang/Nim/issues/19357

advaita-saha: Ohh right 
Will resolve this

constantine/ethereum_verkle_primitives.nim
```diff
@@ -0,0 +1,89 @@
...
+  return check1 and check2
+
+## ############################################################
+##
+##                      Batch Operations
+##
+## ############################################################
+func batchMapToScalarField*[N: static int](
+      res: var array[N, Fr[Banderwagon]], 
+      points: array[N, ECP_TwEdwards_Prj[Fp[Banderwagon]]]): bool {.discardable.} =
```
mratsim, line 62: We might need those to be `openArray` in the future

constantine/ethereum_verkle_primitives.nim
```diff
@@ -0,0 +1,89 @@
...
+  ## on a batch of points
+  ## 
+  ## The batch inversion used in this using
+  ## the montogomenry trick, makes is faster than
+  ## just iterating of over the array of points and 
+  ## converting the curve points to field elements
+  ## 
+  ## Spec : https://hackmd.io/wliPP_RMT4emsucVuCqfHA?view#MapToFieldElement
+  
+  var ys: array[N, Fp[Banderwagon]]
```
mratsim, line 73: In case the input is an openArray, and so size is dynamic, this will need to use the `allocStackArray` strategy

constantine/math/elliptic/ec_twistededwards_batch_ops.nim
```diff
@@ -0,0 +1,115 @@
...
+       affs: var array[N, ECP_TwEdwards_Aff[F]],
+       projs: array[N, ECP_TwEdwards_Prj[F]]) {.inline.} =
+  batchAffine(affs.asUnchecked(), projs.asUnchecked(), N)
+
+func batchAffine*[M, N: static int, F](
+       affs: var array[M, array[N, ECP_TwEdwards_Aff[F]]],
+       projs: array[M, array[N, ECP_TwEdwards_Prj[F]]]) {.inline.} =
+  batchAffine(affs[0].asUnchecked(), projs[0].asUnchecked(), M*N)
+
+func batchInvert_vartime*[N: static int, F](elements: var array[N, F]) =
```
mratsim, line 90: Usually I would implement a 

```Nim
func batchInvert_vartime*[N: static int, F](dst: var array[N, F], src: array[N, F]) =
```

and define the in-place function in terms of that out-of-place function

constantine/math/elliptic/ec_twistededwards_batch_ops.nim
```diff
@@ -0,0 +1,115 @@
...
+  ##  Montgomery's batch inversion
+  ##  This function person the inversion in-place
+  var res: array[N, F]
+  var zeros: array[N, bool]
```
mratsim, line 94: We might need the openArray + allocStackArray there

mratsim (approved):
OK overall,

I would change the `array` input to use openArray instead because:
- This is more flexible, many times the input size isn't known at compile-time
- Code size, this creates a proc per `N` (monomorphization), which isn't worth it for such a slow code.

batchAffine has an example on how to do this properly:
- tag the function `{.noInline.}` so that the allocStack doesn't lead to stack overflow if the function is called in a loop and inlined for some reason (LTO ...)
- if copy from a temporary, use for loop, it will be optimized by the compiler. (alternatively copyMem https://nim-lang.org/docs/system.html#copyMem%2Cpointer%2Cpointer%2CNatural )

advaita-saha:
I have also added the tests to nimble file

mratsim:
Perfect, thanks!

PR #279, opened 2023-10-02
Parallel Ethereum protocols (BLS signature and KZG)

mratsim:
This implements parallel versions of 2 Ethereum protocols
- Batched BLS signature verification
- EIP-4844 KZG polynomial commitment

### Constantine

Clang is significantly faster than GCC for Constantine

image

### Nimbus / nim-blscurve

The BLST backend is mostly agnostic to Clang/GCC choice due to assembly

image

Both Constantine and nim-blscurve use the same parallelization strategy. A significant advantage for Constantine lies in the final exponentiation, however batch verification amortizes its cost over a large batch and somewhat erase the perf difference.

### vs Go/Rust builtin BLST

BLST for Rust and Go uses a different parallelization strategy which would be interesting to bench.

We modify the Rust bench to use 6, 60, 180, and also check that inputs are already deserialized and subgroupchecked
image

Performance:

image

Comparing vs nim-blscurve (which uses the same backend), it's unsure if the batch verification parallelization strategy of BLST is better than nim-blscurve or if it's the difference in threadpool implementation or both.

The difference is that BLST parallelization is guaranteed to only schedule work on non-busy cores, at the price of more *atomic* synchronization, i.e. it's finer-grained. In Constantine we divide into as many chunks as there are threads.

### Constantine

image

#### Single-threaded

image

- Commit is almost 2x slower than Constantine
- Proving is about 1.56x slower than Constantine
- Verifying is 1.20x slower than Constantine

#### Multi-threaded (8-physical cores)

image


mratsim:
Following rework the parallelization of strategy of BLS signatures, we get **superlinear speedups!** And the code is simpler to boot as it reuses the threadpool auto-work-splitter.

**previous**
image

**new**
image

## Explanation

Constantine's Threadpool (and [Weave](https://github.com/mratsim/weave)) provides Lazy Binary Splitting to ensure parallel loops and reductions are multithreaded on an as-needed basis, with backpressure feedback if several threads are busy.

This is key to performance when merging the result of 2 threads is an expensive operation (a multi-Miller-Loop).

- Lazy binary-splitting: a run-time adaptive work-stealing scheduler.
  Tzannes, A., G. C. Caragea, R. Barua, and U. Vishkin.
  In PPoPP ’10, Bangalore, India, January 2010. ACM, pp. 179–190.
  https://user.eng.umd.edu/~barua/ppopp164.pdf

The previous approach (same as nim-blscurve) was eagerly splitting based on the number of cores, not taking into account how busy they were.

The superlinear results are likely due to latency-hiding.

## Impact

Assuming full Ethereum blocks that require verification of 6 signatures each:

- block proposals signatures
- randao reveal signatures
- proposer slashings signatures
- attester slashings signatures
- attestations signatures
- validator exits signatures

We get with 8 cores:
- verifying 1 block, 6 sigs: 1.94x speedup over serial (previous nim-blscurve approach: 1.85x)
- verifying 10 blocks, 60 signatures: 8.32x speedup over serial (previous: 5.63x)
- verifying 30 blocks, 180 signatures: 9.78x speedup over serial (previous: 6.83x)

so 43~48% improvement over nim-blscurve parallel approach.

PR #281, opened 2023-10-15
Allow installation / import through nimble

mratsim:
By default, nimble strips anything not named `constantine` but we need `metering` as well.

This adapt the nimble file to allow this.

Note: the current directory structure was chosen after careful consideration to market across devs looking for different capabilities in cryptographic backend. Metering is a particularly valuable niche to optimize EVM gas costs hence it SHOULD be "in-your-face" in the directory structure.

We might also want to consider installing the binding generators, but it's unsure if there is a command (maybe `nimble build`?) to build Constantine as a static or shared library. Though that might not be needed in the future with incremental compilation.

Lastly we might also consider installing the tests so they can be run on constantine install, it's especially important given unique codepaths for 32/64-bit, CPU ISAs and OS combination.

### DIFF
--- a/constantine.nimble
+++ b/constantine.nimble
@@ -8,6 +8,8 @@ license       = "MIT or Apache License 2.0"
 # ----------------------------------------------------------------
 
 requires "nim >= 1.6.12"
+installDirs "constantine",
+            "metering"
 
 # Nimscript imports
 # ----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


nayr7:
I'm not sure if this is the right place to ask (sorry if it's not) but is there any interest to make `constantine/threadpool` into it's own nimble package like Weave and nim-taskpools ? I am not sure if there are technical hurdles that prevents this  (for example shared code between the threading and cryptographic part) or if it's just a lack of interest. I personally think there might be some merit to be able to use constantine's threadpool without bringing in the all the crypto baggage with it. Also thanks for your work ! :)

PR #287, opened 2023-10-21
Rust bindings

mratsim:
Most provers are written in Rust, so we need to provide Rust bindings for consumption.

In particular we want to provide a backend to test Halo2 KZG's hardware accel API: https://github.com/privacy-scaling-explorations/halo2/issues/216


mratsim:
The PR adds the following:
- Add a ``constantine-sys`` Rust package that exposes BN254-Snarks, BLS12-381 and the Pasta curves as well as Ethereum BLS signatures protocols.
- Add a skeleton of high-level curves and protocol package to expose a Rust API on top of the low-level one.
- Instead of building many libraries we build a single big one, a configuration tool might be provided in the future so that only the required modules are built in. Having multiple libraries would lead to namespacing problem on duplicated functions and status code like serialization or the threadpool.
- Nim+Rust are optimized together through cross-language ThinLTO
- The library is now built with `--passC:-fmerge-all-constants` to help on merging constant arrays that contains things like Montgomery constants.
  The core of duplication problem was addressed by https://github.com/nim-lang/Nim/commit/df7b4bb / https://github.com/nim-lang/RFCs/issues/257#issuecomment-703102705.
  However, it may be possible for Nim to create temporaries in its codegen
  This might also help ICF (Identical Code Folding / Identical COMDAT folding) in C compiler.
- Calling NimMain for CPU feature detection / filling globals is not needed anymore. This should work but is not tested with MSVC.
  - x86 features detection has been cleaned up and focused. 319 LOC instead of 779 LOC.

Issue #244, opened 2023-06-02
Reactivate 32-bit CI


mratsim:
32-bit CI is failing recently. It may be fixed whern a new Github CI Agent is out with more recent packages that avoid the ``you have held broken packages`` error.

![image](https://github.com/mratsim/constantine/assets/22738317/904faa3b-36e5-48dd-9e3c-5a4db5a8ebef)

```
Run sudo dpkg --add-architecture i386
  sudo dpkg --add-architecture i386
  sudo apt-fast update -qq
  
  # Try to fix "E: Unable to correct problems, you have held broken packages."
  sudo apt-fast clean
  
  sudo DEBIAN_FRONTEND='noninteractive' apt-fast install \
    --no-install-recommends -yq \
    gcc-multilib g++-multilib \
    libssl-dev:i386 libgmp-dev:i386
  
  mkdir -p external/bin
  cat << EOF > external/bin/gcc
  #!/bin/bash
  exec $(which gcc) -m32 "\$@"
  EOF
  cat << EOF > external/bin/g++
  #!/bin/bash
  exec $(which g++) -m32 "\$@"
  EOF
  chmod 755 external/bin/{gcc,g++}
  echo '/home/runner/work/constantine/constantine/external/bin' >> $GITHUB_PATH
  shell: /usr/bin/bash -e {0}
[apt-fast 06:24:01]
[apt-fast 06:24:01]Working... this may take a while.
E: Unable to correct problems, you have held broken packages.
[apt-fast 06:24:01]Package manager quit with exit code.
Error: Process completed with exit code 100.
```

mratsim:
Reactivated in #288, and also fuzzed 24/7 in Google OSSfuzz

PR #288, opened 2023-10-21
32-bit fixes

mratsim:
Raised in Google OSS Fuzz

The fixes introduced by #286 didn't compile on 32-bit, this should fix that.

Additionally, we check if #244 was fixed upstream by CI agents with newer packages, and so add back 32-bit tests to CI.

cc @guidovranken


mratsim:
So 32-bit CI can be reactivated.

As a bonus, we can also enable assembly on x86-32 targets (well no one uses those anymore so ....)

PR #292, opened 2023-10-26
Conversion of little endian hex strings to bigints

markspanbroek:
Allows `BigInt.fromHex()` to work with both little-endian and big-endian hex strings.

### DIFF
--- a/constantine/math/io/io_bigints.nim
+++ b/constantine/math/io/io_bigints.nim
@@ -131,13 +131,11 @@ func marshal*(
 #
 # ############################################################
 
-func fromHex*(a: var BigInt, s: string) =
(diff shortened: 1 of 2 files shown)


mratsim (commented):
I'm OK on the technicals but I'm curious what it will be used for.

This is an internal proc for Constantine purposes (as are the BigInts in general, they're here as backend for fields).

And internally or in all protocol specifications big-endian is used.

markspanbroek:
> I'm curious what it will be used for.

In Codex, we're currently working on poseidon2 hashes of byte sequences. We'd like to use little endian encoding to convert bytes to field elements. 

To test this, I'd like to use some example data in unit tests, specified as hex strings. And I'd like to interpret these test strings as little endian encoded field elements.

Issue #302, opened 2023-11-21
Bug: Conditional syntax error for multiplexers.nim


benbierens:
On line 150 and line 167 of multiplexers.nim (sha 5f7ba18)
```
(`*x`)
```
Replaced by
```
(*`x`)
```

mratsim:
What Nim and C compiler are you using, and are you using LTO? Because this doesn't happen in CI and the code has been like this since forever.

mratsim:
closed by #306

PR #306, opened 2023-11-27
Fixes deference in multiplexers.nim

benbierens:


dryajov:
Are there any way of testing this?

mratsim:
See #302 

It's likely correct because the quoting `` is for Nim interpolating its variable into C as mentioned here https://nim-lang.org/docs/manual.html#statements-and-expressions-assembler-statement

```Nim
proc addInt(a, b: int): int =
  asm """
    addl %%ecx, %%eax
    jno 1
    call `raiseOverflow`
    1:
    :"=a"(`result`)
    :"a"(`a`), "c"(`b`)
  """
```

The `*` is for dereferencing the hidden pointer.

Which will need some more special-casing for Nim v2 btw: https://github.com/status-im/nim-stint/blob/711cda4/stint/private/primitives/extended_precision_64bit_uint128.nim#L18-L20

The weird part is that I never triggered an error in years while targeting many platforms in CI.

Issue #303, opened 2023-11-22
Doc: Clarify Intel ASM syntax, LTO, Clang/GCC, MacOS vs Linux/Windows combinations


mratsim:
See https://github.com/codex-storage/nim-codex/pull/625#discussion_r1397703104

Currently Constantine uses Intel ASM syntax because it is LTO proof and works with both GCC and Clang. Except for Apple Clang which misses an upstream fix (Apple bug report FB12137688).

Besides no assembly,
- using AT&T syntax and GCC works
- as does using AT&T syntax and Clang and no LTO

The options should appear in the build system options and pointed to in the README: https://github.com/mratsim/constantine/blob/d77bb79a41b7920505f1915454c802f4d8c1977c/constantine.nimble#L17-L52

Issue #295, opened 2023-11-18
Scoped inline assembly ATT/Intel


mratsim:
Due to incompatibilities with other libraries using AT&T inline assembly like secp256k1 or EVMone or differential fuzzing, it would be best to scope the inline assembly otherwise only GCC can build the combination of Constantine + another lib.

See https://github.com/codex-storage/nim-codex/pull/625#discussion_r1397594279

cc @markspanbroek @dryajov.

This might be possible with `localpassC` pragma with care for assembly that is inlined at another level.
Unsure of what happens with LTO though.

mratsim:
Tried in #309 it doesn't work. Intel syntax and LTO will be disabled by default except for dynamic libraries.

PR #309, opened 2023-12-02
Rework LTO, Intel Assembly syntax and default compile-options

mratsim:
This tries to address #303 #295 https://github.com/codex-storage/nim-codex/pull/625#discussion_r1397703104 following making Constantine compatible with LTO #231

- Intel assembly is used on a per-file basis to avoid interfering with other libraries like secp256k1
- On Apple MacOS/iOS it will need to be explicitly enabled, due to:
  - Clang not supporting Intel syntax:
    - This has not be upstreamed in Apple Clang: https://reviews.llvm.org/D113707
    - Opened issue to Apple: FB12137688
  - Intel Apple CPUs being outdated and not being used for high-performance cryptography and Intel syntax being irrelevant for ARM CPUs.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,7 @@
 [build]
 # https://doc.rust-lang.org/rustc/linker-plugin-lto.html
 rustflags="-Clinker-plugin-lto -Clinker=clang -Clink-arg=-fuse-ld=lld"
+
+# Note, on MacOS, for Intel targets, LTO on the Nim code is not used
+# as Apple Clang does not support the Intel assembly syntax.
+# Unfortunately the AT&T syntax causes issues when LTO inlines constants into inline assembly.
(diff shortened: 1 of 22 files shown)


mratsim:
@dryajov @benbierens @markspanbroek 

Constantine now uses Intel Syntax only if LTO is chosen instead of always, hence it isn't needed to set it in libraries that depend on it.

https://github.com/mratsim/constantine/blob/747830b0c30772b7c2fa61da9d6edb046767712a/constantine/platforms/config.nim#L31-L56

If using LTO:
- Intel syntax is necessary for Clang, recommended for GCC (or it will spam spurious warnings that may denial-of-service the console) and so activated by default.
- Dynamic linking solves assembly syntax incompatibility with other libraries like libsecp256k1
- On MacOS, Intel Syntax requires upstream LLVM Clang, not Apple Clang
- If static linking, using the same compiler and the same major version is strongly recommended

PR #304, opened 2023-11-23
Productionize KZG EIP-4844

mratsim:
This addresses #300

- [x] C API for Ethereum KZG EIP-4844
  - [ ] Parallel C API
  - [ ] Tests for C
- [ ] Go wrapper for fuzzing in https://github.com/jtraglia/kzg-fuzz (cc @jtraglia)
- [ ] Rust wrapper for benchmarking in https://github.com/sifraitech/rust-kzg (cc @sauliusgrigaitis)
- [x] Update trusted setup to official output of KZG ceremony / mainnet: https://github.com/CarlBeek/kzg-ceremony-verifier/blob/master/output_setups/trusted_setup_4096.json
  - [ ] Optional, ability to ingest the official json trusted setup (ideally using a formally verified or correct-by-construction json parser like https://project-everest.github.io/assets/everparse.pdf
  - [x] Or use the c-kzg-4844 format to limit switching friction

PR #318, opened 2023-12-12
Ethereum EIP-4844: Go wrapper for KZG commitments

mratsim:
This is a follow-up of #304.

It adds a go wrapper for Ethereum KZG commitments.

In particular this will allow differential fuzzing vs [c-kzg-4844](https://github.com/ethereum/c-kzg-4844) and [go-kzg-4844](https://github.com/crate-crypto/go-kzg-4844) in [kzg-fuzz](https://github.com/jtraglia/kzg-fuzz) cc @jtraglia.

Some peculiarities:
- On Windows, Go cannot statically link vs `.lib` file directly like C, Nim or Rust. We need to use `-Wl,-Bstatic -lconstantine -Wl,-Bdynamic` as a workaround. Fix raised upstream https://github.com/golang/go/pull/64679.
  Note that that workaround doesn't work on MacOS due to Xcode/Apple Clang not supporting the flag.
- On MacOS, it seems like this fscanf requires an extra byte in the character buffer or it triggers AddressSanitizer or a stack check fail from the fstack-protector: https://github.com/mratsim/constantine/blob/1f3e6b19040d324c7ba25278772b26d1ec0f381b/constantine/trusted_setups/ethereum_kzg_srs.nim#L201-L206

Running go, using the optional `CC=clang` for maximum performance:
```
git clone https://github.com/mratsim/constantine
cd constantine
CC=clang nimble make_lib
cd constantine-go
go test -modfile=../go_test.mod
```

Constantine comes with a separate `go_test.mod` for extra test dependencies. Test dependencies must be specified with `-modfile=../go_test.mod`. The library itself is dependency-free.

For the trusted setup, Constantine supports the same format as c-kzg-4844.


constantine-go/constantine.go
```diff
@@ -0,0 +1,180 @@
...
+		return false, errors.New("VerifyBlobKzgProofBatch: Lengths of inputs do not match.")
+	}
+
+	status := C.ctt_eth_kzg_verify_blob_kzg_proof_batch(
+		ctx.cCtx,
+		*(**C.ctt_eth_kzg_blob)(unsafe.Pointer(&blobs)),
+		*(**C.ctt_eth_kzg_commitment)(unsafe.Pointer(&commitments)),
+		*(**C.ctt_eth_kzg_proof)(unsafe.Pointer(&proofs)),
+		(C.size_t)(len(blobs)),
+		(*C.uint8_t)(unsafe.Pointer(&secureRandomeBytes)),
```
jtraglia, line 171: Minor typo here:
```suggestion
		(*C.uint8_t)(unsafe.Pointer(&secureRandomBytes)),
```

constantine-go/constantine.go
```diff
@@ -0,0 +1,180 @@
...
+	if status != C.cttEthKzg_Success {
+		err := errors.New(
+			C.GoString(C.ctt_eth_kzg_status_to_string(status)),
+		)
+		return false, err
+	}
+	return true, nil
+}
+
+func (ctx EthKzgContext) VerifyBlobKzgProofBatch(blobs []EthBlob, commitments []EthKzgCommitment, proofs []EthKzgProof, secureRandomeBytes [32]byte) (bool, error) {
```
jtraglia, line 159: Also needs to be fixed here:
```suggestion
func (ctx EthKzgContext) VerifyBlobKzgProofBatch(blobs []EthBlob, commitments []EthKzgCommitment, proofs []EthKzgProof, secureRandomBytes [32]byte) (bool, error) {
```

constantine/trusted_setups/ethereum_kzg_srs.nim
```diff
@@ -221,29 +221,29 @@ proc load_ckzg4844(ctx: ptr EthereumKZGContext, f: File): TrustedSetupStatus =
 
   block:
     # G1 points - 96 characters + newline
-    var bufG1Hex {.noInit.}: array[2*g1Bytes, char]
+    var bufG1Hex {.noInit.}: array[2*g1Bytes+1, char] # On MacOS, an extra byte seems to be needed for fscanf or AddressSanitizer complains
```
jtraglia, line 224: This is a good find by the sanitizer. It's right. Because you're reading a string (rather than a byte like it's done in c-kzg-4844) it will add a null terminator at the end.

jtraglia (changes requested):
LGTM 👍 just two small nits.

constantine/trusted_setups/ethereum_kzg_srs.nim
mratsim: I actually open files in binary mode not text: https://github.com/mratsim/constantine/blob/c73ce0e6a0662e905ffc3a670f222818aebe3bfe/constantine/platforms/fileio.nim#L40-L57

And the modifier for strings is %s, which doesn't appear in `const readHexG1 = "%96[^\r\n]%n%*[\r\n]"`
In fact, the sanitizer on Linux doesn't complain.

jtraglia: It shouldn't matter that it's opened in binary mode. It just matters that it's reading a string.

Here's an example with `sscanf` which should operate the same:

```c
#include <stdio.h>

int main(void)
{
    char buffer[97];
    int i, n;
    const char *staticString = \
        "0123456789" // 10
        "0123456789" // 20
        "0123456789" // 30
        "0123456789" // 40
        "0123456789" // 50
        "0123456789" // 60
        "0123456789" // 70
        "0123456789" // 80
        "0123456789" // 90
        "012345";    // 96

    for (i = 0; i < 97; i++) {
        buffer[i] = 0xFF;
    }

    printf("\nInitial buffer contents:");
    for (i = 0; i < 97; i++) {
        if (i % 10 == 0) printf("\n");
        printf("%02X ", (unsigned char)buffer[i]);
    }
    printf("\n");

    sscanf(staticString, "%96[^\r\n]%n%*[\r\n]", buffer, &n);

    printf("\nBuffer contents after sscanf:");
    for (i = 0; i < 97; i++) {
        if (i % 10 == 0) printf("\n");
        printf("%02X ", (unsigned char)buffer[i]);
    }
    printf("\n");

    printf("\nNumber of characters read: %d\n", n);

    return 0;
}
```

For me this is the output:

```
$ make test && ./test
cc  -I/opt/homebrew/include/ -L/opt/homebrew/include/ -L/opt/homebrew/lib/  test.c   -o test

Initial buffer contents:
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF FF FF FF
FF FF FF FF FF FF FF

Buffer contents after sscanf:
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 36 37 38 39
30 31 32 33 34 35 00

Number of characters read: 96
```

As you can see, the 97th byte has been set to 0 but only 96 bytes were read.

jtraglia: > And the modifier for strings is `%s`, which doesn't appear in `const readHexG1 = "%96[^\r\n]%n%*[\r\n]"`
In fact, the sanitizer on Linux doesn't complain.

Oops didn't respond to this. It doesn't need to be `%s` for it to be recognized as a string.

PR #327, opened 2023-12-29
R1CS - Circom parser

mratsim:
Time to tackle proof systems now that Constantine has a robust backend and most Ethereum protocol needs are catered to.

This adds support for R1CS binary format parsing as standardized by circom: https://github.com/iden3/r1csfile/blob/master/doc/r1cs_bin_format.md

Furthermore the PR gathers some constraint / polynomial / proof system IR research in the `research` folder.


mratsim:
Replaced by #388

PR #330, opened 2023-12-29
Add Verkle IPA to constantine.nimble

agnxsh:
added verkle ipa tests to constantine.nimble


mratsim:
I'll have a look on the remaining Windows only crash

image

I suspect, there is a stack-overflow as Windows stack is only 1MB.

There is probably a way to put everything on the heap using ref object.

For example, here I had to change the array when testing KZG to a ref array by allocating it with `new(array[size, byte])`

https://github.com/mratsim/constantine/pull/270/commits/2de92ba73239baff86b0c3792f4bb4522db9e4d6#diff-7ce9bb21f583919ca4dfb0d92f5cba9ac540621843397357f34092f9ca1fb752R87

mratsim:
Can you rebase this with the latest changes?

mratsim:
Some investigation, there seems to be a stack corruption in this PR as `__chkstk_ms` is trigger, see output from running under GDB on Windows via git bash

```
[Suite] Barycentric Form Tests
  [OK] Testing absolute integers
  [OK] Testing Basic Interpolation, without precompute optimisations
  [OK] Divide on Domain using Barycentric Precomputes
  [OK] Testing Barycentric Precompute Coefficients

[Suite] Random Elements Generation and CRS Consistency
  [OK] Test for Generating Random Points and Checking the 1st and 256th point with the Verkle Spec

[Suite] Computing the Correct Vector Commitment
  [OK] Test for Vector Commitments from Verkle Test Vectors by @Ignacio

[Suite] Deserialize a proof which contains an invalid final scalar by @Ignacio
  [OK] Deserialize a proof which contains a final scalar bigger than the field size (MUST fail)

[Suite] Deserialize a proof which wrong lengths by @Ignacio
  [OK] Deserialize a proof which wrong lengths (all MUST fail)

[Suite] Transcript Tests
  [OK] Transcript Testing with different challenge scalars to test randomness
  [OK] Transcript testing with same challenge scalar to test transcript correctness
  [OK] Transcript testing with +1 and -1, appending them to be a compound challenge scalar
  [OK] Transcript testing with point append
  [OK] Transcript testing with repititive append of scalars, thereby a compound challenge scalar

[Suite] IPA proof tests
  [OK] Test for initiating IPA proof configuration
  [OK] Verfify IPA Proof inside the domain by @Ignacio
  [OK] Test for IPA proof consistency
  [OK] Test for IPA Proof of Creation and Verification
  [OK] Test for IPA proof equality

[Suite] Multiproof Tests
  [OK] IPA Config test for Multiproofs

Thread 1 received signal SIGSEGV, Segmentation fault.
___chkstk_ms () at ../../../libgcc/config/i386/cygwin.S:126
126     ../../../libgcc/config/i386/cygwin.S: No such file or directory.
(gdb) bt
#0  ___chkstk_ms () at ../../../libgcc/config/i386/cygwin.S:126
#1  0x0000000000210a38 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

`No such file or directory` just means that gdb cannot display the assembly code as the source assembly is missing.
But __chkstk_ms is the stack canary. It means that:
- either we corrupted a function, this would happen with alloca or playing with pointers, which is unlikely here.
- or we allocated too much on the stack, for example by using a large array.

The good news is that when rebased to latest master, the stack issue disappear, see https://github.com/mratsim/constantine/pull/389.
The bad news is that now the `Test for IPA Proof of Creation and Verification` is failing at line 639.
https://github.com/mratsim/constantine/blob/7ea21a2e45c5e02f8aff2cfc1d94cd3191108559/tests/t_ethereum_verkle_ipa_primitives.nim#L591-L640

For now closing, merging will continue in #389.

PR #331, opened 2024-01-03
[End-user workaround] Constantine-rust: public fields and Default fields

mratsim:
Requested by https://github.com/sifraitech/rust-kzg

Due to the orphan rule preventing building traits on top of external libraries, the Sifraitech team needs the fields to be public.

Unfortunately this conflicts with security best practices and isolation concerns, especially given that Constantine may handle sensitive secrets that must be protected in a best effort manner. Keeping fields private is a good way to avoid accidentally spilling data in a debug log for example.

This exposes the fields in a branch while keeping the library pristine.

This also adds a `Default` trait implementation for convenience. For elliptic curves using it as-is MAY result in undefined behavior for Jacobian or Projective points, as `setInf` is the proper initialization.
This commit _[Ensure the projective point at infinity is not all zeros, but (0, 1, 0)](https://github.com/mratsim/constantine/pull/243/commits/c43919dcf5a5010b541d881f1667e9dc1ba89611)_ fixed an issue with all zero initialization.
Though I'm not sure how to automate it with `bindgen`.

cc @armantidas @sauliusgrigaitis

### DIFF
--- a/constantine-rust/constantine-sys/README.md
+++ b/constantine-rust/constantine-sys/README.md
@@ -1,0 +1,4 @@
+# Constantine-sys (Experimental)
+
+This is an experimental version of Constantine-Rust with public fields
+For use in https://github.com/sifraitech/rust-kzg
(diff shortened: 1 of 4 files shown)


sauliusgrigaitis:
Thanks, if you want to work on `rust-kzg` update then check https://github.com/ArtiomTr/rust-kzg/tree/Integrate_sppark_msm . This is going to be merged soon. It also contains the updated performance graphs. Seems `constantine` peformance penalty is gone!

mratsim:
Happy New Year,

I've updated the bindings to latest Constantine.

PR #334, opened 2024-01-07
make_lib_rust - use position independent code

lynxcs:
added -fPIC when using make_lib_rust, since [rust already does that by default.](https://github.com/rust-lang/rust/pull/16340)

This change is required by [rust-kzg constantine backend](https://github.com/sifraitech/rust-kzg/pull/251) to make c-kzg-4844 bindings work.

@mratsim if enabling for all rust builds is not acceptable, I can also add as optional feature in Cargo.toml & change how it's passed in build.rs

I am not familiar with Nim so if there's a better/more preferred way of doing this lmk, will be happy to fix.

### DIFF
--- a/constantine.nimble
+++ b/constantine.nimble
@@ -228,12 +228,13 @@ proc genDynamicLib(outdir, nimcache: string) =
   else:
     compile "libconstantine.so"
 
-proc genStaticLib(outdir, nimcache: string) =
+proc genStaticLib(outdir, nimcache: string, extFlags = "") =
   proc compile(libName: string, flags = "") =
     echo &"Compiling static library:  {outdir}/" & libName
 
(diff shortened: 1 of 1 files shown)


mratsim (dismissed):
Thank you!

I would have done something similar.

mratsim:
Seems like on Windows, `-fPIC` is not needed or even is an error with MinGW (what I use in CI):
- https://mingw-w64-public.narkive.com/4JUWn3PX/position-independent-code-fpic-on-x86-64-windows-dll-specially-for-cygwin
- https://github.com/jedisct1/libsodium/issues/14
- https://github.com/mia-0/rubberband/pull/2

mratsim (approved):

Issue #336, opened 2024-01-08
Add test using verkle-test-vectors for computing the correct Pedersen Commitment


agnxsh:
Add tests to compute the correct Pedersen Commitment for this scalar dataset provided by [Ignacio from the Geth team](https://github.com/jsign/verkle-test-vectors/blob/main/crypto/001_vector_commitment.json).

PR #337, opened 2024-01-08
[aborted] vector commitment test, crs fix and consistency check

### LINKED ISSUE #335: Fix `generate_random_points( )` function to get the correct first and 256th Banderwagon Point.
The existing code for `generate_random_points( )` for Banderwagon does some successive calls `deserialize`, `serialize` and `deserialize`, in an incorrect way, as they are Banderwagon specific functions, hence, it differs from the spec and fails to check with the 1st and 256th point. 

Reference to the incorrect snippet:
https://github.com/mratsim/constantine/blob/df9034b9b451fa156929878fab89ecaf1f4880c9/constantine/eth_verkle_ipa/common_utils.nim#L46-L53

After enough digging, it stands clear that the first `serialize` and `deserialize` calls as per the spec neither checks the mod order p, nor it checks for the Banderwagon subgroup. It does a simple conversion to byte array and the other way round, of just the `x-coordinate` of the point, which in Constantine can be achieved using the `unmarshal` & `marshal` function, with the _Endianness_ set to `bigEndian`.

Ideally the algorithm goes this way (specifc to what we have in Constantine) :
- First we think of a seed, in our case it is `eth_verkle_2021`
- Then we use a cryptographic hash function (sha256 here), to hash the seed, update the hash with an incrementer value initially set to 0, and finalize the hash in a byte array
- Then we interpret this byte array into the $x-coordinate$ of a Banderwagon point, and update the incrementer value by 1.
- We then convert this $x-coordinate$ to a bigEndian byte array once again
- If bigints are needed in between as accumulator variables, then we use `matchingBigInt(Banderwagon)` as the type and NOT `matchingOrderBigInt` as the curve order is dealt with in the scalar field `Fr`.
- We deserialize the bigEndian byte array using Banderwagon's `deserialize` function, this checks 2 extra things :
1 ) taking the _lexicographically smaller_ value for $y$
2 ) checking whether the point generated is in the Banderwagon subgroup
- If these checks pass, we add the point to the list of Banderwagon points comprising of the CRS.

This would get us the correct first and last Banderwagon points, which in turn becomes crucial in computing the correct Pedersen Hash.

agnxsh:
Fixes #335 and #336
The following things are done in this PR:
1. Fix of `generate_random_points( )` function to the get consistent CRS values.
2. Adding vector commitment test vectors from [Ignacio's verkle-test-vectors (by the Geth team)](https://github.com/jsign/verkle-test-vectors/blob/main/crypto/001_vector_commitment.json).


tests/t_ethereum_verkle_ipa_primitives.nim
```diff
@@ -42,106 +47,160 @@ var generator = Banderwagon.getGenerator()
...
+      doAssert arr_byte[255].toHex() == "0x3de2be346b539395b0c0de56a5ccca54a317f1b5c80107b0802af9a62276a4d8", "Failed to generate the 256th point!"
+    
+    testGenPoints()
+
+# ############################################################
+#
+#      Test for Computing the Correct Vector Commitment
+#
+# ############################################################
+
```
advaita-saha, line 180: Please add the link to Ignacio's test data as a comment

advaita-saha (commented):
LGTM

tests/t_ethereum_verkle_ipa_primitives.nim
advaita-saha: Like this 
https://github.com/mratsim/constantine/blob/df9034b9b451fa156929878fab89ecaf1f4880c9/tests/t_ethereum_verkle_primitives.nim#L183-L185

agnxsh: Okay

advaita-saha (dismissed):

mratsim (commented):
Ideally the `1. Indentation fix for eth_verkle_ipa and test files` should have been a separate PR to ease reviews because it all green/red.

What was the issue/fix of 2. regarding consistent CRS?

PR #350, opened 2024-01-23
add: IPA Proof and Multiproof Ser/De

### LINKED ISSUE #349: IPAProof and Multiproof Ser/De
An important feature of stateless clients is to send over and receive the _Execution Witness_ (part of which is`VerkleProofs`) via the network. The `VerkleProofs` consists of the `Multiproof` which the client requires to peform _vector commitment openings_ to compute the correct path to a given key-value entry in the Verkle Tree. This idea is similar to the _Merkle path_ concept of usual Merkle Trees.

Additionally, cross testing it with a valid Multiproof data used by another client implementation is also appreciated. One such valid test vector is [this](https://github.com/crate-crypto/go-ipa/blob/master/multiproof_test.go#L120-L121) 

Finally, upstreaming the required types and low level functions are also to be done, to provide a layer of abstraction to the trie-implementers/client devs.

agnxsh:
Fixes #349 

IPA Proofs in Verkle consists of 

1) a Left array of 8 Base Field points, a Right array of 8 Base Field points, 

2) and a Scalar Field element. 

During serialization the format goes as follows:

`L[0] (32 - byte array) L[1] (32 - byte array) .... L[7] (32 - byte array) ..... R[0] (32 - byte array) ... R[7] (32 - byte array) A (32 - byte array)`
 
 Which means the size of the byte array should be : 
 
`32 * 8 (for Left half) + 32 * 8 (for Right half) + 32 * 1 (for Scalar) = 32 * 17 = 544 elements in the byte array.`

- Note that checks like Subgroup check for Banderwagon Points for Base Field elements in L and R, and checks for a valid scalar checking the Banderwagon Curve Order is mandatory. They are all checked in the further low level functions

- Note that check for Lexicographically Largest criteria for the Y - coordinate of the Twisted Edward Banderwagon point is also mandatory and, is pre-checked within this function from the `deserialize` function.

Multiproofs in Verkle have a format of 

1) The queried Base Field where the Vector Commitment `opening` is created, consider this as the equivalent to the `Merkle Path` in usual Merkle Trees.

2) The entire IPAProof which is exactly a 576 byte array, go through `serializeIPAProof` for the breakdown
 
The format of serialization is as:

`Query Point (32 - byte array) .... IPAProof (544 - byte array) = 32 + 544 = 576 elements in the byte array`


mratsim (approved):
Thanks!

PR #354, opened 2024-02-02
precomp sqrt optimization

### LINKED ISSUE #236: Faster point decompression
cc @ec2.

Point decompression is a bottleneck in protocols (or to load trusted setups).

It's very slow due to square root.

image

Storing point uncompressed would be a memcopy instead (something like 20000x faster).

This is worse for non-sqrt friendly curves like BLS12-377 or Bandersnatch.

See:
- Tonelli-Shanks with precomputed dlog tables\
   https://hackmd.io/@jsign/bandersnatch-optimized-sqrt-notes
   https://ihagopian.com/posts/tonelli-shanks-with-precomputed-dlog-tables
- Computing Square Roots Faster than the Tonelli-Shanks/Bernstein Algorithm\
  Palash Sarkar\
  https://eprint.iacr.org/2020/1407

advaita-saha:
# Optimization of square roots Tonelli-Shanks, with pre-computed dlog tables

Fixes #236 

Notes about the approach : 
 - https://ihagopian.com/posts/tonelli-shanks-with-precomputed-dlog-tables
 - https://hackmd.io/@jsign/bandersnatch-optimized-sqrt-notes

Reference Implementation from Gottfried in `gnark`
https://github.com/GottfriedHerold/Bandersnatch/blob/f665f90b64892b9c4c89cff3219e70456bb431e5/bandersnatch/fieldElements/field_element_square_root.go

#### Currently pre-computes are added for Bandersnatch & Banderwagon


constantine/math/elliptic/ec_twistededwards_affine.nim
```diff
@@ -121,7 +122,10 @@ func trySetFromCoordX*[F](P: var ECP_TwEdwards_Aff[F], x: F): SecretBool =
   P.x += one
 
   # √((1 - ax²)/(1 - dx²))
-  result = sqrt_ratio_if_square(t, P.x, P.y)
+  var temp {.noInit.}: F
+  temp.inv(P.y)
+  temp.prod(P.x, temp)
+  result = t.sqrtPrecomp(temp) # TODO: needs to be made conditional
```
mratsim, line 128: `sqrtPrecomp` is only defined for Banderwagon / Bandersnatch and will break for Curve25519 for example.
Furthermore it is not constant-time and may handle secrets, so it needs to be explicitly suffixed with vartime as well as its callers.

Instead `invsqrt_vartime` should be created with special handling that uses the dlog for Banderwagon / Bandersnatch, similar to:
https://github.com/mratsim/constantine/blob/5894a8da757bb935b681c73d1fdc92a7838f94ed/constantine/math/arithmetic/finite_fields_square_root.nim#L198-L214
And a new `trySetFromCoordX_vartime` should be used.

constantine/math/constants/banderwagon_sqrt.nim
```diff
@@ -17,3 +18,1344 @@ const
...
+      Fp[Banderwagon].fromHex"0x5e73ed8c432405d0ae25b21df6c52b2a6c7876bf0928e68fae34a81e1cfca2ec",
+      Fp[Banderwagon].fromHex"0x6a945980dade66efbcca660ee327eb377a1cb8c070f0ce322e9a4db5f138dab3",
+      Fp[Banderwagon].fromHex"0x1f3834a0f387add7faa1f3804fdb73d008ccffd43f1931b63ff630d3f89c4c2c",
+      Fp[Banderwagon].fromHex"0x1f2caa7aeef7d8dee0eb9d91cdecccfbbf5354ab6e8c2ab4f007bdc150d024b1",
+      Fp[Banderwagon].fromHex"0x640d097461a2eeaf4e84b3cd7dc75b61db872ef3dce28e788d28c143168bba2c",
+      Fp[Banderwagon].fromHex"0x36e1ebe124834a8fcfdc1da7bf59004b42f464cb4256a3b3f3f90bd2b05e602e",
+      Fp[Banderwagon].fromHex"0x60c7efec34093e9a725ce48f79bf9cd60081066e578d78bf6207cd2f20df9d12",
+      Fp[Banderwagon].fromHex"0x18ba7658084c71a467e701a945e9b713df7b4ee66e4caccb7d01866add958780"
+    ]
+  ]
```
mratsim, line 1361: Note for future improvement, a Sage script generates this table as this can benefit BLS12-377 and the Pasta curves

benchmarks/bench_verkle_primitives.nim
```diff
@@ -0,0 +1,144 @@
...
+#   else:
+#     var P, Q: Prj
+#     P.fromAffine(Banderwagon.getGenerator())
+#     Q.fromAffine(Banderwagon.getGenerator())
+#   bench("Banderwagon Equality ", T, iters):
+#     assert (P == Q).bool()
+
+
+
+proc serializaBench*(T: typedesc, iters: int) =
```
mratsim, line 115:
```suggestion
proc serializeBench*(T: typedesc, iters: int) =
```

mratsim (commented):
typo

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -0,0 +1,198 @@
...
+  negExponent = negExponent shr 1
+  z.setOne()
+
+  for i in 0..<Fp.C.tonelliShanks(sqrtParam_Blocks):
+    sqrtAlg_GetPrecomputedRootOfUnity(temp, int((negExponent shr (i*Fp.C.tonelliShanks(sqrtParam_BlockSize))) and Fp.C.tonelliShanks(sqrtParam_BitMask)), uint(i))
+    z.prod(z, temp)
+
+  return true
+
+func sqrtPrecomp*(dst: var Fp, x: Fp): SecretBool {.inline.} =
```
mratsim, line 185: This function is not necessary, `invSqrt` is the more universal primitive and what should be exported

mratsim:
Some more comments: I think we can totally remove the `if` checks inside the algorithm.

1. This facilitates constant time implementation
2. checking if the number is a square is easy once we have computed the candidate `sqrt`, we can just multiply it by itself, which is somewhat cheap, and it is what we do: https://github.com/mratsim/constantine/blob/5894a8da757bb935b681c73d1fdc92a7838f94ed/constantine/math/arithmetic/finite_fields_square_root.nim#L244-L257
3. The case that matters and does NOT need constant-time is deserialization, and wrong deserialization means some protocol is not followed properly which means we need to stop interacting with the other party (if network) or the database (if local). This is useful for batch deserialization to shave 10~100 microseconds over thousands of points, but failing earlier after 5us instead of 10us is not necessary. (even for a DOS we're not in the milliseconds range here)

mratsim:
I can confirm a 50% perf improvement on my machine :fire: 🎉

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -0,0 +1,198 @@
...
+  return 0
+  
+# sqrtAlg_GetPrecomputedRootOfUnity sets target to g^(multiplier << (order * sqrtParam_BlockSize)), where g is the fixed primitive 2^32th root of unity.
+#
+# We assume that order 0 <= order*sqrtParam_BlockSize <= 32 and that multiplier is in [0, 1 <<sqrtParam_BlockSize)
+func sqrtAlg_GetPrecomputedRootOfUnity*(target: var Fp, multiplier: int, order: uint) =
+  target = Fp.C.tonelliShanks(sqrtPrecomp_PrecomputedBlocks)[order][multiplier]
+
+
+func sqrtAlg_ComputerRelevantPowers*(z: var Fp, squareRootCandidate: var Fp, rootOfUnity: var Fp) =
```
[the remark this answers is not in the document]
mratsim: Actually computing acc should be a private function that uses the `{.addchain.}` macro which will use an accelerated square/prod for addition chains: for example https://github.com/mratsim/constantine/blob/5894a8da757bb935b681c73d1fdc92a7838f94ed/constantine/math/constants/bls12_381_sqrt.nim#L19-L21

https://github.com/mratsim/constantine/blob/5894a8da757bb935b681c73d1fdc92a7838f94ed/constantine/math/arithmetic/finite_fields.nim#L560-L592

advaita-saha:
I will be solving the constant time issues
I wanted to make sure first that if the optimisation is working as expected

mratsim:
Strangely I can't find the sqrt or deserialization benches in go-ipa.

advaita-saha:
> Strangely I can't find the sqrt or deserialization benches in go-ipa.

There isn't, I am pushing in a few hours 
Only bench that ignacio did on my request was of serialisation 
https://github.com/crate-crypto/go-ipa/pull/62

advaita-saha:
Benchmarking code for `go-ipa`
https://github.com/advaita-saha/go-ipa/blob/15686fb2ed7bf62b2c9883b17c77e3de1d09be0a/banderwagon/element_test.go#L393-L409

agnxsh:
> I will be solving the constant time issues
> 
> I wanted to make sure first that if the optimisation is working as expected

Also if it's necessary I can give a benchmark on M2 Pro by coming Monday/Tuesday

advaita-saha:
## Current Benchmarks Tested in AWS instance with AMD EPYC 7R13 Processor

Screenshot 2024-02-07 at 1 56 16 AM

# Results 
w.r.t `go-ipa` **~23%** performance improvement 🔥
w.r.t `constantine(prev-impl)` **~50%** performance improvement 🚀

mratsim:
Confirmed on my machine:

go-ipa:
image

Constantine:
image

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -24,20 +24,19 @@ import
 # The returned value is only meaningful modulo 1<<sqrtParam_BlockSize and is fully reduced, i.e. in [0, 1<<sqrtParam_BlockSize )
 #
 # NOTE: If x is not a root of unity as asserted, the behaviour is undefined.
-func sqrtAlg_NegDlogInSmallDyadicSubgroup*(x: Fp): int =
+func sqrtAlg_NegDlogInSmallDyadicSubgroup(x: Fp): int =
   let key = cast[int](x.mres.limbs[0] and SecretWord 0xFFFF)
-  if key in Fp.C.tonelliShanks(sqrtPrecomp_dlogLUT):
-    return Fp.C.tonelliShanks(sqrtPrecomp_dlogLUT)[key]
-  return 0
+  return Fp.C.sqrtDlog(dlogLUT).getOrDefault(key, 0)
```
mratsim, line 29: This unfortunately isn't constant-time, see: https://github.com/nim-lang/Nim/blob/57658b685cd08dc5a0c1f7a9aa58fafa553efc4e/lib/pure/collections/tableimpl.nim#L187-L191

```Nim
template getOrDefaultImpl(t, key, default: untyped): untyped =
  mixin rawGet
  var hc: Hash
  var index = rawGet(t, key, hc)
  result = if index >= 0: t.data[index].val else: default
```
There is an if-branch.

To be fully constant-time, and ensure it stays that way, Constantine cannot depend on the standard library for code that may handle secrets.

Here, we can instead rename to `sqrtAlg_NegDlogInSmallDyadicSubgroup_vartime` and suffix all callers with vartime. I'll try to see later if we can have a constant-time LUT, or maybe split it into 2 LUTs.

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -24,20 +24,19 @@ import
 # The returned value is only meaningful modulo 1<<sqrtParam_BlockSize and is fully reduced, i.e. in [0, 1<<sqrtParam_BlockSize )
 #
 # NOTE: If x is not a root of unity as asserted, the behaviour is undefined.
-func sqrtAlg_NegDlogInSmallDyadicSubgroup*(x: Fp): int =
+func sqrtAlg_NegDlogInSmallDyadicSubgroup(x: Fp): int =
```
mratsim, line 27:
```suggestion
func sqrtAlg_NegDlogInSmallDyadicSubgroup_vartime(x: Fp): int {.tags:[VarTime], raises: [].} =
```

There might be errors that might ask you to add other effects or mention unlisted exceptions

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -148,51 +147,44 @@ func invSqrtEqDyadic*(z: var Fp): bool =
   # set powers[i] to z^(1<< (i*blocksize))
   var powers: array[4, Fp]
   powers[0] = z
-  for i in 1..<Fp.C.tonelliShanks(sqrtParam_Blocks):
+  for i in 1..<Fp.C.sqrtDlog(Blocks):
```
mratsim, line 150:
```suggestion
  for i in 1 ..< Fp.C.sqrtDlog(Blocks):
```

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -148,51 +147,44 @@ func invSqrtEqDyadic*(z: var Fp): bool =
...
     powers[i] = powers[i - 1]
-    for j in 0..<Fp.C.tonelliShanks(sqrtParam_BlockSize):
+    for j in 0..<Fp.C.sqrtDlog(BlockSize):
```
mratsim, line 152:
```suggestion
    for j in 0 ..< Fp.C.sqrtDlog(BlockSize):
```

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -148,51 +147,44 @@ func invSqrtEqDyadic*(z: var Fp): bool =
...
-  if (negExponent and 1) == 1:
-    return false
+  # if (negExponent and 1) == 1:
+  #   return false
+
+  # result = SecretBool((negExponent and 1) != 1)
 
-  for i in 1..<Fp.C.tonelliShanks(sqrtParam_Blocks):
-    temp2 = powers[Fp.C.tonelliShanks(sqrtParam_Blocks) - 1 - i]
+  for i in 1..<Fp.C.sqrtDlog(Blocks):
```
mratsim, line 168:
```suggestion
  for i in 1 ..< Fp.C.sqrtDlog(Blocks):
```

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -148,51 +147,44 @@ func invSqrtEqDyadic*(z: var Fp): bool =
...
     var newBits = sqrtAlg_NegDlogInSmallDyadicSubgroup(temp2)
-    negExponent = negExponent or (newBits shl ((i*Fp.C.tonelliShanks(sqrtParam_BlockSize)) - Fp.C.tonelliShanks(sqrtParam_FirstBlockUnusedBits)))
+    negExponent = negExponent or (newBits shl ((i*Fp.C.sqrtDlog(BlockSize)) - Fp.C.sqrtDlog(FirstBlockUnusedBits)))
 
   negExponent = negExponent shr 1
   z.setOne()
 
-  for i in 0..<Fp.C.tonelliShanks(sqrtParam_Blocks):
-    sqrtAlg_GetPrecomputedRootOfUnity(temp, int((negExponent shr (i*Fp.C.tonelliShanks(sqrtParam_BlockSize))) and Fp.C.tonelliShanks(sqrtParam_BitMask)), uint(i))
+  for i in 0..<Fp.C.sqrtDlog(Blocks):
```
mratsim, line 180:
```suggestion
  for i in 0 ..< Fp.C.sqrtDlog(Blocks):
```

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -148,51 +147,44 @@ func invSqrtEqDyadic*(z: var Fp): bool =
...
-  sqrtAlg_ComputerRelevantPowers(xCopy, candidate, rootOfUnity)
-  if not invSqrtEqDyadic(rootOfUnity):
-    return SecretBool(false)
-
+  sqrtAlg_ComputeRelevantPowers(x, candidate, rootOfUnity)
+  invSqrtEqDyadic(rootOfUnity)
   dst.prod(candidate, rootOfUnity)
-  return SecretBool(true)
-
+  dst.inv()
```
mratsim, line 190: Assuming the input `x`

The result of `invSqrtEqDyadic(rootOfUnity)`

should be `x⁻¹ᐟ²` according to https://github.com/GottfriedHerold/Bandersnatch/blob/7c0464f7dfae8f1139c9f812d1543de631f5262b/bandersnatch/fieldElements/field_element_square_root.go#L273-L277
```go
// invSqrtEqDyadic asserts that z is a 2^32 root of unity and tries to set z := 1/sqrt(z).
//
// If z is actually a 2^32th *primitive* root of unity, the square root does not exist and we return false without modifying z.
// Otherwise, z is changed to 1/sqrt(z) and we return true
func (z *feType_SquareRoot) invSqrtEqDyadic() (ok bool) {
```

then the last two lines do

```
dst <- candidate * x⁻¹ᐟ²
dst <- dst⁻¹ (and this is equal to x⁻¹ᐟ² because it matches with invsqrt)
```

Ergo, `candidate = x` and `rootOfUnity = x⁻¹ᐟ²` before the last 2 lines, so they are unnecessary.

This can be raised as a refactoring issue and then cleaned up afterwards.

constantine/math/arithmetic/finite_fields_square_root.nim
```diff
@@ -210,6 +211,8 @@ func invsqrt*[C](r: var Fp[C], a: Fp[C]) =
     r.invsqrt_p3mod4(a)
   elif C.has_P_5mod8_primeModulus():
     r.invsqrt_p5mod8(a)
+  elif C == Bandersnatch or C == Banderwagon:
+    r.inv_sqrt_precomp(a)
```
mratsim, line 215: We will need this in a new separate function called `invsqrt_vartime`

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
mratsim, line 1: The file should be renamed `finite_fields_square_root_precomp.nim so it's next to the other file when listing.

constantine/math/elliptic/ec_twistededwards_affine.nim
```diff
@@ -10,7 +10,7 @@ import
   ../../platforms/abstractions,
   ../config/curves,
   ../arithmetic,
-  ../arithmetic/finite_fields_precomp_square_root,
+  #../arithmetic/finite_fields_precomp_square_root,
```
mratsim, line 13:
```suggestion
```

mratsim (commented):
The most important thing is to rename the `sqrtAlg_NegDlogInSmallDyadicSubgroup` function with the _vartime suffix as well as its callers.

The functions in finite_fields_square_root.nim starting from `invsqrt` should be duplicated to have a `..._vartime` suffix and `trySetFromCoordX` should have a `trySetFromCoordX_vartime` version as well.

This is important so that protocols that use invsqrt or sqrt with secret data do not suddenly leak secrets.

The deserialization functions should then point to that `vartime` function.

Plan for the future:
1. Add SageCode to generate the precomputation automatically
2. Expand that precomputation to other curves with no fast sqrt like the pasta curves and BLS12-377
3. Refactor the algorithm to make it constant-time and remove the dependency on Nim standard-library.

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
advaita-saha: I have seen this while implementing, that's why I haven't commented this particular function.
The problem is when I try to reverse this from the back, then it doesn't work which implies `x != candidate`, which actually should be

But again if `sqrtX = candidate * x⁻¹ᐟ²`, then candidate should be equal to `x`. I have been confused initially with this, but left this to proceed with the implementation

mratsim: I see, then we can leave it as is and scheduled that as a next refactoring.

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
```diff
@@ -0,0 +1,190 @@
...
+  ../../platforms/abstractions,
+  ../constants/zoo_square_roots,
+  ./bigints, ./finite_fields
+
+# ############################################################
+#
+#            Only For Bandersnatch/Banderwagon for now
+#
+# ############################################################
+
```
mratsim, line 20: We need credits to Gottfried implementation here

mratsim (commented):
---
For the current state:

Also @GottfriedHerold, it seems like your code has no license, can you clarify whether a direct port in Constantine (license MIT+Apache) is OK.


----

For a future PR

After thinking a bit, in a future PR, to enable this optimization for:
- constant-time square root (for example hash-to-curve)
- other curves
- remove Nim std/tables dependency (due to exceptions, there is no alloc since it's compile-time tables)

We will need:
- smaller tables, ideally configurable, because 65536 size is too big for constant-time
- a Sagemath code to generate the LUT tables.

As such it makes sense to reuse Pornin's:
- Optimized Discrete Logarithm Computation for Faster Square Roots in Finite Fields\
  Thomas Pornin, 2023\
  https://eprint.iacr.org/2023/828
- https://github.com/pornin/modsqrt/blob/main/modsqrt.sage

for the constant-time implementation. And it's probably possible to get it within 5% of Gottfried algorithm for the variable-time so we can reuse the same tables.

constantine/math/arithmetic/finite_fields_precomp_square_root.nim
mratsim: A similar trick to Tonelli-Shanks as described here can probably be applied:

https://github.com/mratsim/constantine/blob/5894a8da757bb935b681c73d1fdc92a7838f94ed/sage/square_root_bls12_377.sage#L222-L248

mratsim:
It probably would be cleaner to rebase on top of master instead of merging master into this branch to keep a linear history.

GottfriedHerold:
#Also @GottfriedHerold, it seems like your code has no license, can you clarify whether a direct port in Constantine (license MIT+Apache) is OK.

Yes, that's OK.

constantine/math/arithmetic/finite_fields_square_root.nim
```diff
@@ -210,9 +211,14 @@ func invsqrt*[C](r: var Fp[C], a: Fp[C]) =
     r.invsqrt_p3mod4(a)
   elif C.has_P_5mod8_primeModulus():
     r.invsqrt_p5mod8(a)
+  elif C == Bandersnatch or C == Banderwagon:
+    r.inv_sqrt_precomp_vartime(a) # should be changed
```
mratsim, line 215: This should be completely removed

constantine/math/arithmetic/finite_fields_square_root.nim
```diff
@@ -210,9 +211,14 @@ func invsqrt*[C](r: var Fp[C], a: Fp[C]) =
...
   else:
     r.invsqrt_tonelli_shanks(a)
 
+func invsqrt_vartime*[C](r: var Fp[C], a: Fp[C]) =
+  r.inv_sqrt_precomp_vartime(a)
```
mratsim, line 220:
```suggestion
  ## Compute the inverse square root of ``a``
  ##
  ## This requires ``a`` to be a square
  ##
  ## The result is undefined otherwise
  ##
  ## The square root, if it exist is multivalued,
  ## i.e. both x² == (-x)²
  ## This procedure returns a deterministic result
  ## This procedure is NOT constant-time
  when C.has_P_3mod4_primeModulus():
    r.invsqrt_p3mod4(a)
  elif C.has_P_5mod8_primeModulus():
    r.invsqrt_p5mod8(a)
  elif C == Bandersnatch or C == Banderwagon:
    r.inv_sqrt_precomp_vartime(a)
  else:
    r.invsqrt_tonelli_shanks(a)
```

constantine/math/arithmetic/finite_fields_square_root_precomp.nim
```diff
@@ -0,0 +1,193 @@
...
+  acc.square_repeated(8)
+  acc.prod(acc, z255)
+  acc.square_repeated(8)
+  acc.prod(acc, z255)
+  acc.square_repeated(8)
+  acc.prod(acc, z255)
+  # acc is now z^((BaseFieldMultiplicativeOddOrder - 1)/2)
+  rootOfUnity.square(acc)
+  rootOfUnity *= z
+  squareRootCandidate.prod(acc, z)
```
mratsim, line 137: For future reference, I have a feeling that this addchain computes the same thing as https://github.com/mratsim/constantine/blob/973e0aaf9b3a7ce185d30750afc8ee12dc042cd6/constantine/math/constants/bandersnatch_sqrt.nim#L19-L149

Also it would need to be moved to Banderwagon/Bandersnatch specific files in the future

constantine/math/constants/bandersnatch_sqrt.nim
```diff
@@ -5,19 +5,11 @@
...
 
-const
-  # with e = 2adicity
-  # p == s * 2^e + 1
-  # root_of_unity = smallest_quadratic_nonresidue^s
-  # exponent = (p-1-2^e)/2^e / 2
-  Bandersnatch_TonelliShanks_exponent* = BigInt[222].fromHex"0x39f6d3a994cebea4199cec0404d0ec02a9ded2017fff2dff7fffffff"
-  Bandersnatch_TonelliShanks_twoAdicity* = 32
-  Bandersnatch_TonelliShanks_root_of_unity* = Fp[Bandersnatch].fromHex"0x212d79e5b416b6f0fd56dc8d168d6c0c4024ff270b3e0941b788f500b912f1f"
-
```
mratsim, line 20: This should stay here

constantine/math/constants/bandersnatch_sqrt.nim
```diff
@@ -155,3 +147,1353 @@ func precompute_tonelli_shanks_addchain*(
   r *= x11111111
   r.square()
   r *= a
+
```
mratsim, line 150:
```suggestion

# ############################################################
#
#       Optimized square-root via Discrete Log lookup tables
#
# ############################################################

```

constantine/math/constants/bandersnatch_sqrt.nim
```diff
@@ -155,3 +147,1353 @@ func precompute_tonelli_shanks_addchain*(
...
+const
+  # with e = 2adicity
+  # p == s * 2^e + 1
+  # root_of_unity = smallest_quadratic_nonresidue^s
+  # exponent = (p-1-2^e)/2^e / 2
+  Bandersnatch_TonelliShanks_exponent* = BigInt[222].fromHex"0x39f6d3a994cebea4199cec0404d0ec02a9ded2017fff2dff7fffffff"
+  Bandersnatch_TonelliShanks_twoAdicity* = 32
+  Bandersnatch_TonelliShanks_root_of_unity* = Fp[Bandersnatch].fromHex"0x212d79e5b416b6f0fd56dc8d168d6c0c4024ff270b3e0941b788f500b912f1f"
```
mratsim, line 158: This should stay above the Specialized Tonelli-SHanks section

constantine/math/constants/banderwagon_sqrt.nim
```diff
@@ -18,6 +19,1347 @@ const
   Banderwagon_TonelliShanks_twoAdicity* = 32
   Banderwagon_TonelliShanks_root_of_unity* = Fp[Banderwagon].fromHex"0x212d79e5b416b6f0fd56dc8d168d6c0c4024ff270b3e0941b788f500b912f1f"
 
+  Banderwagon_SqrtDlog_TotalBits* = Banderwagon_TonelliShanks_twoAdicity
```
mratsim, line 22: This part should be below the `Specialized Tonelli-Shanks` section

and with a title

```Nim
# ############################################################
#
#       Optimized square-root via Discrete Log lookup tables
#
# ############################################################
```

constantine/math/constants/banderwagon_sqrt.nim
```diff
@@ -154,4 +1496,4 @@ func precompute_tonelli_shanks_addchain*(
   # 261 + 3 = 264 operations
   r *= x11111111
   r.square()
-  r *= a
+  r *= a
```
mratsim, line 1499: Files should end with empty-lines, normally vscode and vim/nvim do that automatically

constantine/serialization/codecs_banderwagon.nim
```diff
@@ -123,7 +123,7 @@ func deserialize_unchecked*(dst: var EC_Prj, src: array[32, byte]): CttCodecEccS
   var x{.noInit.}: Fp[Banderwagon]
   x.fromBig(t)
 
-  let onCurve = dst.trySetFromCoordX(x)
+  let onCurve = dst.trySetFromCoordX_vartime(x) # later to be shifted to a constant time version
```
mratsim, line 126:
```suggestion
  let onCurve = dst.trySetFromCoordX_vartime(x) # later to be shifted to a constant time version
```

deserialized_unchecked should be duplicated into:
- deserialized_unchecked_vartime, and the "# later to be shifted to a constant time version" is unnecessary
- deserialized_unchecked, that uses the regular `trySetFromCoordX`

Then if the constant-time is fast enough, we can deprecate the vartime version to keep the API simple,
if it's too slow, we can maintain both.

mratsim (approved):
LGTM, some nits.

constantine/math/arithmetic/finite_fields_square_root.nim
advaita-saha: Ahh just missed my eye

advaita-saha:
@mratsim did the suggested changes

benchmarks/bench_verkle_primitives.nim
```diff
@@ -0,0 +1,144 @@
...
+    discard P.deserialize(bytes)
+
+
+proc main() =
+  # separator()
+  # equalityBench(Aff, Iters)
+  # equalityBench(Prj, Iters)
+  separator()
+  serializeBench(Prj, Iters)
+  deserializeBench(Prj, Iters)
```
mratsim, line 141: We probably need `deserialize_vartime` and `deserialize_unchecked_vartime` benches now

mratsim (approved):
LGTM.

I can either merge as-is or wait for the extra benches to be added.

PR #356, opened 2024-02-11
Square Root: addition chain for Bandersnatch and banderwagon

mratsim:
This implements a low-hanging fruit addition chain for square root for Bandersnatch and Banderwagon so that #354 is benched against fully optimized Tonelli-Shanks.

cc @advaita-saha


mratsim:
Improvement is around 5%

image

image

PR #364, opened 2024-02-22
Verkle IPA: Backend maintenance and further testing

agnxsh:
This is a maintenance PR which changes a couple of things:

- Added a `populateCoefficientVector( )`function to handle cases where the evaluation point is Inside/Outside Verkle Domain separately
- Added a Transcript test to check consistency of challenge scalar after pointAppend
- Added a test to ensure in domain division is working perfectly (using Barycentric Precomputes) and cross-tested it with Go
- Added test [011](https://github.com/jsign/verkle-test-vectors/blob/main/crypto/011_range_proof_verification.json) and [012](https://github.com/jsign/verkle-test-vectors/blob/main/crypto/012_in_domain_ipa_proof_verification.json) from Go verkle-test-vectors, and made fixes in functions wherever needed. The verkle cryptography backend is now fully consistent with Go and Nethermind, and can be further used in Verkle Tries without correctness issues (most likely) :) 
- Added test vectors from Nethermind team to ensure IPA proof consistency
- Moved `evalOutsideDomain()` to common utilities because it can be used in Verkle Tries for testing
- Upstreamed some of the leftover APIs and fixed indentation of that file.

Further possible optimizations:
- Parallelize IPA Proof creation
- Parallelize evaluating multiple polynomials

cc: thanks to @tanishqjasoria and @jsign for helping me with the test vectors.


constantine/eth_verkle_ipa/barycentric_form.nim
```diff
@@ -29,7 +29,7 @@ func newPrecomputedWeights* [PrecomputedWeights] (res: var PrecomputedWeights) =
 
   var midpoint = VerkleDomain
   for i in 0 ..< midpoint:
-    var weights {.noInit.}: Fr[Banderwagon]
+    var weights: Fr[Banderwagon]
```
mratsim, line 32: Is this related to the random garbage you mentioned on Discord https://discord.com/channels/613988663034118151/1138059404831379508/1210186486662565888?

computeBarycentricWeights below is always initialized so it's strange that it fixes uninitialized memory: https://github.com/mratsim/constantine/blob/c7979b003372b329dd450ff152bb5945cafb0db5/constantine/eth_verkle_ipa/barycentric_form.nim#L62-L81

constantine/eth_verkle_ipa/barycentric_form.nim
```diff
@@ -136,30 +136,29 @@ func getBarycentricInverseWeight*(res: var Fr[Banderwagon], precomp: Precomputed
...
-    res = x
+     res = x
 
 
-func divisionOnDomain*(res: var array[VerkleDomain,Fr[Banderwagon]], precomp: PrecomputedWeights, index:  var int, f:  openArray[Fr[Banderwagon]]) =
+func divisionOnDomain*(res: var array[VerkleDomain,Fr[Banderwagon]], precomp: PrecomputedWeights, index:  var uint8, f: openArray[Fr[Banderwagon]]) =
   ## Computes f(x) - f(x_i) / x - x_i using the barycentric weights, where x_i is an element in the
   var is_negative = true
-  var y = f[index]
+  var y = f[int(index)]
```
mratsim, line 151:
```suggestion
  let index = int(index)
  var y = f[index]
```
is simpler but why use uint8 in the first place?

Also this file needs `{. raises: [].}` and `{.push checks:off.}` to avoid Nim introducing exceptions and allocations for them.

constantine/eth_verkle_ipa/common_utils.nim
```diff
@@ -91,33 +91,34 @@ func computeInnerProducts* [Fr] (res: var Fr, a,b : View[Fr])=
...
+  doAssert a.len == b.len , "Should have equal lengths!"
 
   for i in 0 ..< a.len:
     var bx {.noInit.}: EC_P
-
-    b[i].scalarMul(x.toBig())
     bx = b[i]
-    res[i].sum(bx, a[i])
+    var x_big {.noInit.}: matchingOrderBigInt(Banderwagon)
+    x_big = x.toBig()
```
mratsim, line 111:
```suggestion
    x_big.fromField(x)
```
or keep as before.
Alternatively a `scalarMul` overload that accepts field elements

constantine/eth_verkle_ipa/ipa_prover.nim
```diff
@@ -48,93 +49,125 @@ func genIPAConfig*(res: var IPASettings, ipaTranscript: var IpaTranscript[Crypto
...
+    a_copy[i] = a[i]
+ 
+  var q {.noInit.}: EC_P
   q = ic.Q_val
   q.scalarMul(w)
 
-  var current_basis {.noInit.}: array[VerkleDomain, EC_P]
-  current_basis = ic.SRS
+  # var current_basis {.noInit.}: array[VerkleDomain, EC_P]
+  # current_basis = ic.SRS
```
mratsim, line 112: leftover comments?

constantine/eth_verkle_ipa/ipa_prover.nim
```diff
@@ -48,93 +49,125 @@ func genIPAConfig*(res: var IPASettings, ipaTranscript: var IpaTranscript[Crypto
...
+
+  var cb_c = newSeq[EC_P](VerkleDomain)
```
mratsim, line 114: I avoid using sequences and Nim-allocated code in general.
`allocHeapArray` should be preferred.

The previous view approach completely avoids allocations and is better.

constantine/eth_verkle_ipa/ipa_verifier.nim
```diff
@@ -30,7 +30,7 @@ func generateChallengesForIPA*(res: var openArray[matchingOrderBigInt(Banderwago
     transcript.pointAppend( asBytes"R", proof.R_vector[i])
     res[i].generateChallengeScalar(transcript,asBytes"x")
 
-func checkIPAProof* (ic: IPASettings, transcript: var CryptoHash, commitment: var EC_P, proof: IPAProof, evalPoint: Fr[Banderwagon], res: Fr[Banderwagon]) : bool = 
+func checkIPAProof* (got: var EC_P, ic: IPASettings, transcript: var CryptoHash, commitment: var EC_P, proof: IPAProof, evalPoint: Fr[Banderwagon], res: Fr[Banderwagon]) : bool =
```
mratsim, line 33: argument order is (context, result, the rest), similar to secp256k1: https://github.com/bitcoin-core/secp256k1/blob/0653a25d50f67c68bd2d196ecc7eddab067d95ef/include/secp256k1.h#L14-L17

constantine/eth_verkle_ipa/multiproof.nim
```diff
@@ -59,39 +51,33 @@ func computePowersOfElem*(res: var openArray[Fr], x: Fr, degree: SomeSignedInt)=
...
-  var Cs_prime {.noInit.} : array[VERKLE_DOMAIN, EC_P]
-  for i in 0 ..< VERKLE_DOMAIN:
-    Cs_prime[i] = Cs[i]
-
   for i in 0 ..< num_queries:
-    transcript.pointAppend(asBytes"C", Cs_prime[i])
-    var z {.noInit.} : Fr[Banderwagon]
-    z.domainToFrElem(Zs[i])
+    transcript.pointAppend(asBytes"C", Cs[i])
+    var z: Fr[Banderwagon]
```
mratsim, line 73: The noInit should still work

constantine/eth_verkle_ipa/multiproof.nim
```diff
@@ -265,81 +240,87 @@ func verifyMultiproof*(multiProof: var MultiProof, transcript : var CryptoHash,
...
-    for i in 0 ..< VERKLE_DOMAIN:
-      var z {.noInit.}: Fr[Banderwagon]
-      z.domainToFrElem(uint8(i))
-
-      helperScalarDeno[i].diff(t_fr, z)
+    groupedEvals[z] += scaledEvals
 
-    var helperScalarDeno_prime: array[VERKLE_DOMAIN, Fr[Banderwagon]]
-    helperScalarDeno_prime.batchInvert(helperScalarDeno)
+  #Calculating the helper scalar denominator, which is 1 / t - z_i
```
mratsim, line 260:
```suggestion
  # Calculating the helper scalar denominator, which is 1 / t - z_i
```

constantine/eth_verkle_ipa/transcript_gen.nim
```diff
@@ -31,6 +31,10 @@ func newTranscriptGen*(res: var CryptoHash, label: openArray[byte]) =
 func messageAppend*(res: var CryptoHash, message: openArray[byte], label: openArray[byte]) =
   res.update(label)
   res.update(message)
+  # var hash {.noInit.}: array[32, byte]
+  # res.finish(hash)
+  # res.update(hash)
```
mratsim, line 36: leftovers?

constantine/eth_verkle_ipa/barycentric_form.nim
agnxsh: Nope this wasn't the place, but yes the final value of `weights` shouldn't change any of the values as we're iterating through the `VerkleDomain`, ok i can revert it back.

constantine/eth_verkle_ipa/barycentric_form.nim
agnxsh: Uint8 just to hard ensure the input is restricted to `mod` 256, and now when I change it back to `int` i get the moded value incase there was an input greater than the VerkleDomain

constantine/eth_verkle_ipa/ipa_prover.nim
agnxsh: So here, apart from chunking `a` to `a_L` and `a_R`, and so on, I need to cut the length of `a` itself by half, during every iteration, and the IPA goes on for 8 rounds. I think when I was using `View`, probably the original array wasn't shrinking to half. Also `splitMiddle` was earlier used in `StridedView` in `kzg/research` I think but it didn't have a `toOpenArray( )` method at that point, so I went with a rather basic approach. After every iteration, if `allocHeapArray` would let me shrink the array after every iteration, then I'll definitely implement that. Could you please point out such an example?

okay I can do `View` with `chunk`, but that again can work if there's a reallocating mechanism, else it's sometimes gives me correctness issues.

mratsim (approved):
LGTM,

Future cleanups:

- there is a `newSeq` introduced that needs to be replaced to avoid dependency on Nim runtime, especially if we want to expose IPA through C/Go/Rust
- some proc like absIntChecker are exported but might be local-only
- scalarMul needs an overload for field element and not just big ints
- evalInDomain / evalOutside domain and the Lagrange form of IPA are duplicates of KZG the polynomial functions need to be deduplicated and parallelized.
- views need a splitMiddle functions and StridedView can then be replaced

Issue #359, opened 2024-02-12
Sage script for dlog precomp constants calculation 


advaita-saha:
The PR #354 adds the pre-comp optimisation for `sqrt`, which has the constants prepared only for 
 - Banderwagon Curve
 - Bandersnatch Curve

A `sage` script is supposed to be added for the calculation of the these constants for other curves. The algorithm for the calculation of the constants can be found here 
https://github.com/crate-crypto/go-ipa/blob/408dbffb2041271c95979a3fb79d98b268bf2880/bandersnatch/fp/sqrt.go#L45-L89

A corresponding sage script is supposed to be created and added inside the sage folder

rupam-04:
@mratsim As far as I know, sage only supports Weierstrass curves and doesn't support Twisted Edwards curves like Banderwagon and Bandersnatch. So is there any way around this?

mratsim:
> @mratsim As far as I know, sage only supports Weierstrass curves and doesn't support Twisted Edwards curves like Banderwagon and Bandersnatch. So is there any way around this?

We're working at the field level so we only need
```sage
Fp = GF('0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001')
```

mratsim:
Pasting my comment on modularization https://github.com/mratsim/constantine/pull/369#issuecomment-2094811256

> One thing though that I think is important is this:
> 
>     * https://github.com/pornin/modsqrt/blob/0a9c69a0ce8b4a031e7f4a728a3bd93870088ecf/modsqrt.sage#L133-L138
> 
>     * https://github.com/pornin/modsqrt/blob/0a9c69a0ce8b4a031e7f4a728a3bd93870088ecf/modsqrt.sage#L232-L236
> 
> 
> ```python
> #  - When lb is small enough, this is a matter of recognizing the value z
> #    among the known 2^lb-th roots of 1, with a "reverse lookup". This can
> #    be done by using only a small number of bits from the value (see the
> #    minhc() function in the code below, to find a minimal-sized bit
> #    pattern that is enough to distinguish all roots). We can thus stop
> #    the recursion before reaching lb == 1.
> ```
> and the function minhc: https://github.com/pornin/modsqrt/blob/0a9c69a0ce8b4a031e7f4a728a3bd93870088ecf/modsqrt.sage#L535-L559
>
[...]

Issue #358, opened 2024-02-12
precomp square root in constant time


advaita-saha:
# The square root with precomp optimisation is not currently in constant-time

PR #354 adds the precomp optimisation to the square-root for Banderwagon/Bandersnatch.
The implementation needs to be changed to a constant-time implementation and all the `_vatime` functions to be changes/removed related to the optimised square root. 

Once the implementation is in constant-time, a check for the curve type should be added here and the precomp sqrt to be called
https://github.com/mratsim/constantine/blob/5894a8da757bb935b681c73d1fdc92a7838f94ed/constantine/math/arithmetic/finite_fields_square_root.nim#L244-L257

## The problem which need to be fixed
So the problem is in the access of the LUT table in `constantine/math/constants/banderwagon_sqrt.nim` and `constantine/math/constants/bandersnatch_sqrt.nim`. We need to fetch values corresponding to the key, but if the key is not present we need to return zero. This is currently done using the `tableLUT.getOrDefault(key, 0)`. But this is function is not in constant time. 
If this access is made in constant-time then the whole implementation will be in constant-time

rupam-04:
I came up with a quite simple solution to this issue, but I am not sure if its actually in constant time. While reading the code I came across a certain key value pair in the table (65534: 0). The problem was the time for accessing a value in a table with its key was not the same as returning 0. We can kind of avoid this issue with the following implementation:

```nim
func sqrtAlg_NegDlogInSmallDyadicSubgroup(x: Fp): int {.tags:[VarTime], raises: [KeyError].} =
  let key = cast[int](x.mres.limbs[0] and SecretWord 0xFFFF)
  if Fp.C.sqrtDlog(dlogLUT).hasKey(key):
    return Fp.C.sqrtDlog(dlogLUT)[key]
  else:
    return Fp.C.sqrtDlog(dlogLUT)[65534]
```

Since in both the cases of the `if` we are accessing a table element, it is supposed to be in constant time I think. So this might be  a workaround for now.

mratsim:
Unfortunately this is still susceptible to cache-timing attacks, like Percival: https://koclab.cs.ucsb.edu/teaching/cren/project/2010/gallegos.pdf

![image](https://github.com/mratsim/constantine/assets/22738317/8545b5ff-0139-4c04-b33d-306dfed3ce9c)

The data in `Fp.C.sqrtDlog(dlogLUT)[65534]` would be read often and be hot in cache and so would result in timing differences compared to a "real" key.

[...]

PR #369, opened 2024-04-22
Sage script for constants calculation for banderwagon and bandersnatch

rupam-04:
Addresses #359 .

Implemented the sqrtPrecomp_PrimitiveDyadicRoots in Sage. However I would need some help on defining the ret array in sage as I didn't come across any helpful SageMath docs.


rupam-04:
Replaced the array with hardcoded variables. While the script seems to be running correctly, I would like some insights on whether this approach is feasible or not.

mratsim:
> However I would need some help on defining the ret array in sage as I didn't come across any helpful SageMath docs.

For arrays, you can get inspiration from my lattice pretty printer here:
https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/sage/derive_endomorphisms.sage#L45-L51

which generates this:

https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/constants/bls12_381_endomorphisms.nim#L21-L27

or my Frobenius constants generator

https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/sage/derive_frobenius.sage#L139-L149

which generates this

https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/constants/bls12_381_frobenius.nim#L51-L129

After the Dyadic roots, we also need: the Precomputed Blocks and the table:

https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/constants/banderwagon_sqrt.nim#L210-L1507

rupam-04:
> > However I would need some help on defining the ret array in sage as I didn't come across any helpful SageMath docs.
> 
> For arrays, you can get inspiration from my lattice pretty printer here:
> 
> https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/sage/derive_endomorphisms.sage#L45-L51
> 
> which generates this:
> 
> https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/constants/bls12_381_endomorphisms.nim#L21-L27
> 
> or my Frobenius constants generator
> 
> https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/sage/derive_frobenius.sage#L139-L149
> 
> which generates this
> 
> https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/constants/bls12_381_frobenius.nim#L51-L129
> 
> After the Dyadic roots, we also need: the Precomputed Blocks and the table:
> 
> https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/constants/banderwagon_sqrt.nim#L210-L1507

I tried using a dictionary instead of an array. That shouldn't be a problem right?

sage/square_root_banderwagon_and_bandersnatch.sage
```diff
@@ -0,0 +1,33 @@
...
+# function sqrtPrecomp_PrecomputedBlocks:
+block = {}
+print('\n\nPrecomputed Blocks:\n')
+for i in range (0, 4):
+    block[i] = {}
+    block[i][0] = 1
+    for j in range (1, 256):
+        block[i][j] = Fp(block[i][j-1] * sqrtPrecomp_PrimitiveDyadicRoots[i * 8])
+        print(hex(block[i][j]))
+
```
advaita-saha, line 30: Print in such a manner such that `i=0`, `i=1`, ..... is visible 
```suggestion
block = {}
print('\n\nPrecomputed Blocks:\n')
for i in range (0, 4):
    block[i] = {}
    block[i][0] = 1
    print("For i = " + i)
    for j in range (1, 256):
        block[i][j] = Fp(block[i][j-1] * sqrtPrecomp_PrimitiveDyadicRoots[i * 8])
        print(hex(block[i][j]))

```

sage/square_root_banderwagon_and_bandersnatch.sage
```diff
@@ -0,0 +1,33 @@
...
+# function sqrtPrecomp_dlogLUT:
```
advaita-saha, line 31: For the implementations you need to convert a point into limbs
Limbs are of 64bits, you can achieve this by 

Assuming that this is related to the Jubjub Edwards curve, for which `d = -(10240/10241) (mod r = 52435875175126190479447740508185965837690552500527637822603658699938581184513) = 19257038036680949359750312669786877991949435402254120286184196891950884077233`, the four 64-bit limbs (in little-endian order) are:

`(73851820219580081, 2967167836563676454, 17725484271033745364, 3067823152860638024)`

The "process" is pretty simple:

- `73851820219580081 = d mod 2^64`
- `2967167836563676454 = (d / 2^64) mod 2^64`
- `17725484271033745364 = (d / 2^128) mod 2^64`
- `3067823152860638024 = (d / 2^192) mod 2^64`

Hope this will help in your implementation, as limbs are not there in sage by default AFAIK

sage/square_root_banderwagon_and_bandersnatch.sage
```diff
@@ -0,0 +1,34 @@
...
+    print(hex(sqrtPrecomp_PrimitiveDyadicRoots[i+1]))
+
+
+# function sqrtPrecomp_PrecomputedBlocks:
+block = {}
+print('\n\nPrecomputed Blocks:\n')
+for i in range (0, 4):
+    block[i] = {}
+    block[i][0] = 1
+    print("For i = " + i)
```
advaita-saha, line 27:
```suggestion
    print("For i = " + str(i))
```

sage/square_root_banderwagon_and_bandersnatch.sage
```diff
@@ -0,0 +1,50 @@
...
+sqrtPrecomp_dlogLUT = {}
+
+rootOfUnity = int(1)
+
+print('\n\nsqrtPrecomp_ReconstructionDyadicRoot = ' + hex(sqrtPrecomp_ReconstructionDyadicRoot) + '\n')
+
+for i in range(LUTSize):
+    mask = LUTSize - 1
+    minus_i = -i % sqrtPrecomp_ReconstructionDyadicRoot
+    sqrtPrecomp_dlogLUT[rootOfUnity & 0xFFFF] = int(minus_i & mask)
```
advaita-saha, line 45: Wrong, please refer the go counterpart
And my previous comment on limbs

advaita-saha (changes requested):
Implementation for LUT is wrong

advaita-saha (changes requested):
Please match your output with expected results. It's not matching

sage/square_root_banderwagon_and_bandersnatch.sage
mratsim: Just using the field printer as hex is fine, Constantine can convert hex to fields at compile-time and it's easier to debug.

mratsim:
> I tried using a dictionary instead of an array. That shouldn't be a problem right?

Using a table is fine at the moment for proof-of-concept purposes. However, they cannot be constant-time and so will need to be changed later: https://github.com/mratsim/constantine/issues/358

rupam-04:
> > I tried using a dictionary instead of an array. That shouldn't be a problem right?
> 
> Using a table is fine at the moment for proof-of-concept purposes. However, they cannot be constant-time and so will need to be changed later: #358

I thought the sage script is just for generating the constants. So is it necessary to convert the sage code to constant time?

sage/square_root_banderwagon_and_bandersnatch.sage
```diff
@@ -0,0 +1,55 @@
...
+rootOfUnity_limbs[1] = 6378425256633387010
+rootOfUnity_limbs[2] = 11064306276430008309
+rootOfUnity_limbs[3] = 1739710354780652911
+rootOfUnity = 0
+
+for i in range(len(rootOfUnity_limbs)):
+    rootOfUnity += rootOfUnity_limbs[i] * (2^64)^i
+
+print('\n\nsqrtPrecomp_ReconstructionDyadicRoot = ' + hex(sqrtPrecomp_ReconstructionDyadicRoot) + '\n')
+
```
advaita-saha, line 49:
```suggestion
print("Banderwagon_SqrtDlog_dlogLUT : ")
```

advaita-saha (changes requested):
LGTM in terms of accuracy of the constants generated.

Should this be modularised such that constants for other curves can also be generated by plugging in the curve params ? @mratsim

sage/square_root_banderwagon_and_bandersnatch.sage
advaita-saha: Yes
But this was for generating the LUT

Check this line :  https://github.com/crate-crypto/go-ipa/blob/408dbffb2041271c95979a3fb79d98b268bf2880/bandersnatch/fp/sqrt.go#L80

mratsim:
> > > I tried using a dictionary instead of an array. That shouldn't be a problem right?
> > 
> > 
> > Using a table is fine at the moment for proof-of-concept purposes. However, they cannot be constant-time and so will need to be changed later: #358
> 
> I thought the sage script is just for generating the constants. So is it necessary to convert the sage code to constant time?

Not yet, but it will need to be refactored for #358.

> Should this be modularised such that constants for other curves can also be generated by plugging in the curve params ? @mratsim 

Yes,  this can be done in a subsequent PR, if this one is too big, atm it's only 47 lines but missing all the code that format things correctly so that we can just copy-paste the constants into the Nim codebase.

One thing though that I think is important is this:
- https://github.com/pornin/modsqrt/blob/0a9c69a0ce8b4a031e7f4a728a3bd93870088ecf/modsqrt.sage#L133-L138
- https://github.com/pornin/modsqrt/blob/0a9c69a0ce8b4a031e7f4a728a3bd93870088ecf/modsqrt.sage#L232-L236

```sage
#  - When lb is small enough, this is a matter of recognizing the value z
#    among the known 2^lb-th roots of 1, with a "reverse lookup". This can
#    be done by using only a small number of bits from the value (see the
#    minhc() function in the code below, to find a minimal-sized bit
#    pattern that is enough to distinguish all roots). We can thus stop
#    the recursion before reaching lb == 1.
```

and the function minhc: https://github.com/pornin/modsqrt/blob/0a9c69a0ce8b4a031e7f4a728a3bd93870088ecf/modsqrt.sage#L535-L559

Ergo, we are using a LUT of window 8 (see also https://hackmd.io/@jsign/bandersnatch-optimized-sqrt-notes#Detailed-comparison):
https://github.com/mratsim/constantine/blob/d7871d74852b360aba456cee710389b7858cc5ac/constantine/math/constants/banderwagon_sqrt.nim#L167-L172

but we're lucky because 8-bit patterns can uniquely identify the root-of-unity we need.

For constant-time implementation because we can't use branches, we have to scan through a whole array that maps bit pattern to root of unities, i.e. no sparse hash-table. And the mapping size grows exponentially with the number of bits, so we likely would want to use 6-bit windows for example if the pattern is small enough to uniquely identify the root of unity we need, that would be 4x less memory consumption.

This means we need to check the absence of pattern collisions in our precomputed small windows.

rupam-04:
@mratsim things are pretty much completed it seems to me. Are we ready to merge this then?

advaita-saha (approved):

mratsim:
Closing this as it's been ~9 months.

Currently unlike all other sage scripts that generate constants, the input cannot be easily copy-pasted into a Nim file as-is. This is key. Afterwards it ideally addresses #358.

PR #370, opened 2024-04-27
EC: Scalar mul accepts scalar as field element as well as bigint

mratsim:
Previously Constantine required to convert scalar input of scalar mul to bigint every-time before calling.

This does not refactor eth-verkle-ipa. It will be done in a second time.


mratsim:
It seems like there is a regression in Nim 2.0.x devel branch that breaks type-checking:

image

No issue in Nim 2.0.2

image

mratsim:
Investigation

Error
```
Error: type mismatch
Expression: isOne((coords: [(mres: (limbs: [0'u, 0'u, 0'u, 0'u, 0'u, 0'u])), (mres: (limbs: [
    16517514583386313282'u, 74322656156451461'u, 16683759486841714365'u,
    815493829203396097'u, 204518332920448171'u, 1306242806803223655'u]))]))
  [1] (coords: [(mres: (limbs: [0'u, 0'u, 0'u, 0'u, 0'u, 0'u])), (mres: (limbs: [
    16517514583386313282'u, 74322656156451461'u, 16683759486841714365'u,
    815493829203396097'u, 204518332920448171'u, 1306242806803223655'u]))]): QuadraticExt

Expected one of (first mismatch at [position]):
[1] func isOne(a: BigInt): SecretBool
[1] func isOne(a: ExtensionField): SecretBool
[1] func isOne(a: FF): SecretBool
```

The input should match:

https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/extension_fields/towers.nim#L124-L129

and ExtensionField is just a generic typeclass

https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/extension_fields/towers.nim#L34-L48

The only potential buggy part is that it's called in a compile-time context:
https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/math/extension_fields/towers.nim#L1442-L1447

PR #371, opened 2024-04-27
EC eth-verkle-ipa: refactor scalar-mul calls following #370

mratsim:
ℹ️ Stacked PR: merge #370 first then change the target branch of this PR to master.

This PR propagates the change of #370 to eth-verkle-ipa and also switch from scalarMul to scalarMul_vartime everywhere.

A followup should be to also do the same for MSM.

cc @agnxsh @advaita-saha


agnxsh:
Looks OK to me

Issue #372, opened 2024-04-27
MacOS CI: Now running on ARM64 CPUs


mratsim:
It seems like assembly for Apple Clang changed:

https://github.com/mratsim/constantine/actions/runs/8862799707/job/24336160802?pr=370

```
/Users/runner/work/constantine/constantine/constantine/nimcache/libcurves_headers/@m..@sconstantine@splatforms@sisa@scpudetect_x86.nim.c:80:8: error: invalid output constraint '=a' in asm
      :"=a"(result.eax), "=b"(result.ebx), "=c"(result.ecx), "=d"(result.edx)
       ^
```
https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/platforms/isa/cpudetect_x86.nim#L17-L36

mratsim:
It seems like the MacOS CI is now using ARM CPUs and x86 assembly code is still called there:

![image](https://github.com/mratsim/constantine/assets/22738317/bdb5e892-618b-4cbc-8723-97da4996749f)

mratsim:
Cross-compiling for Linux ARM64 on Linux x86 host works with Clang.

However looking at the error message we can see `Nim Compiler version 1.6.21 [MacOSX: amd64]`
but the host is actually `arm64-apple-darwin23.4.0`

So I suspect Nim is actually running through Rosetta and "thinks" it is on a x86 platform, which leads to this `const X86` check being mistakenly true:
https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/platforms/config.nim#L12-L13

PR #374, opened 2024-04-28
CI: Github MacOS now uses ARM64. 

mratsim:
See #372

Github runners now use ARM64 instead of x86 for MacOS.

Constantine failed to compile for inline assembly when the target CPU architecture is not explicit.

This suggests that the wrong architecture (x86) was used for compiling on ARM64.

Looking at `nim -v`, Nim x86 was installed and used, probably transparently through Rosetta. And the x86 check at https://github.com/mratsim/constantine/blob/976c8bb215a3f0b21ce3d05f894eb506072a6285/constantine/platforms/config.nim#L12-L13
use the "native" compiler config.

Hence this only affects the CI, and this should be fixable with just a change to arm64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,11 @@ jobs:
             BACKEND: ASM
 
           - os: macos
-            cpu: amd64
+            cpu: arm64
             TEST_LANG: c
             BACKEND: NO_ASM
           - os: macos
-            cpu: amd64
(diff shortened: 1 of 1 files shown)


mratsim:
This is pending Nim MacOS ARM nightly releases: https://github.com/nim-lang/nightlies/pull/82

mratsim:
superceded by #524

PR #377, opened 2024-05-06
update dependency for halo2 MsmAccel trait

Vindaar:
Replaces the dependency for the MSM acceleration layer to use the version from the Privacy Scaling Explorations repo (https://github.com/privacy-scaling-explorations/halo2/)

Still needs to depend on `halo2curves` explicitly, as `halo2_middleware` does not reexport it (and currently I didn't manage to figure out how to fix that).

### DIFF
--- a/constantine-rust/constantine-halo2-zal/Cargo.toml
+++ b/constantine-rust/constantine-halo2-zal/Cargo.toml
@@ -11,7 +11,6 @@ repository = "https://github.com/mratsim/constantine"
 constantine-sys = { path = "../constantine-sys" }
 constantine-core = { path = "../constantine-core" }
 halo2_middleware = { git = 'https://github.com/privacy-scaling-explorations/halo2' }
-halo2curves = { version = "0.6.1", default-features = false }
 
 [dev-dependencies]
 ark-std = "0.3"
--- a/constantine-rust/constantine-halo2-zal/Cargo.toml
+++ b/constantine-rust/constantine-halo2-zal/Cargo.toml
@@ -10,7 +10,8 @@ repository = "https://github.com/mratsim/constantine"
 [dependencies]
 constantine-sys = { path = "../constantine-sys" }
 constantine-core = { path = "../constantine-core" }
-halo2curves = { git = 'https://github.com/taikoxyz/halo2curves', branch = "pr-pse-exec-engine" }
+halo2_middleware = { git = 'https://github.com/privacy-scaling-explorations/halo2' }
+halo2curves = { version = "0.6.1", default-features = false }
 
 [dev-dependencies]
 ark-std = "0.3"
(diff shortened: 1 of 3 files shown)


constantine-rust/constantine-halo2-zal/Cargo.toml
```diff
@@ -22,7 +23,7 @@ rand_xorshift = "0.3"
 rayon = "1.8"
 
 # In CI "asm" needs to be disabled as some agents don't support ADX.
-halo2curves = { git = 'https://github.com/taikoxyz/halo2curves', branch = "pr-pse-exec-engine" }
+# halo2curves = { git = 'https://github.com/taikoxyz/halo2curves', branch = "pr-pse-exec-engine" }
```
Vindaar, line 26: Not sure what should happen with this here.

constantine-rust/constantine-halo2-zal/Cargo.toml
```diff
@@ -10,7 +10,8 @@ repository = "https://github.com/mratsim/constantine"
 [dependencies]
 constantine-sys = { path = "../constantine-sys" }
 constantine-core = { path = "../constantine-core" }
-halo2curves = { git = 'https://github.com/taikoxyz/halo2curves', branch = "pr-pse-exec-engine" }
+halo2_middleware = { git = 'https://github.com/privacy-scaling-explorations/halo2' }
+halo2curves = { version = "0.6.1", default-features = false }
```
Vindaar, line 14: Do we want to depend explicitly on `0.6.1`?

mratsim: Actually the way cargo works, this means a dependency on 0.6.x: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

image

But I think the elliptic curves and CurveAffine should be reexported by Zal or Middleware.

constantine-rust/constantine-halo2-zal/Cargo.toml
mratsim: Leftover from when I enabled multicore and asm features for benching (and disabled it in CI) See: [231ccc7#diff-86e33129ff3e1c7ef4e097a64b3bba9f931352847b95fa0d7c9033665c10b8b6R26](https://github.com/mratsim/constantine/commit/231ccc7acea7975267a49e88e50035b137539b61#diff-86e33129ff3e1c7ef4e097a64b3bba9f931352847b95fa0d7c9033665c10b8b6R26)

It can be deleted with.

Note that if upstream is changed to reexport part of halo2curves, we might need it again for the benchmarks:

https://github.com/mratsim/constantine/blob/12f568654e7b6d7b9b738c4bd3d615932bde8d47/constantine-rust/constantine-halo2-zal/benches/msm.rs#L18-L21

at least the best_multi_exp part.

constantine-rust/constantine-halo2-zal/Cargo.toml
Vindaar: Ah, that clears up a lot, thanks!

mratsim:
This can be updated now that https://github.com/privacy-scaling-explorations/halo2/pull/323 is merged upstream

Vindaar:
Removed the explicit `halo2curves` dependency and we now use it via `halo2_middleware`. I didn't change the commented out dev dependency, because I'm not quite sure as to what extent we still need the alternative `halo2curves`.

mratsim (approved):
Thanks!

PR #380, opened 2024-05-08
Bandersnatch/banderwagon endomorphism acceleration

mratsim:
This implements:

- [x] endomorphism acceleration support for Twisted Edwards curves
- [x] endomorphism acceleration for Bandersnatch and Banderwagon
- [ ] benchmarks for scalar multiplication and MSM

### λ

The endomorphism is equivalent to a scalar multiplication by a solution of λ² = -2.
There are 2 such solutions and for a generic implementation we should try
to ensure which of λ₀ = √-2 and λ₁ = -√-2 corresponds to the computation at https://github.com/mratsim/constantine/blob/69b8f41b71be92ae8bf096b79878338c40446180/constantine/math/constants/zoo_endomorphisms.nim#L49-L73

This is what is done for cubic root endomorphisms:
https://github.com/mratsim/constantine/blob/f3e08ed5a46426990fb5973c23e8e86332137456/sage/derive_endomorphisms.sage#L151-L162
https://github.com/mratsim/constantine/blob/f3e08ed5a46426990fb5973c23e8e86332137456/sage/derive_endomorphisms.sage#L84-L112

However, SageMath has no builtin support for Edwards curves so we would would need to reimplement scalar multiplication to do these checks. Hence we hardcode the λ from the paper.

### Lattice decomposition

The lattice decomposition also follows the reference implementation, i.e. instead of https://github.com/mratsim/constantine/blob/f3e08ed5a46426990fb5973c23e8e86332137456/sage/derive_endomorphisms.sage#L61-L67
we do https://github.com/asanso/Bandersnatch/blob/e280929/python-ref-impl/bandersnatch.py#L20-L21
```python
        M = Matrix([[-L,1], [r,0]])
        self.N = M.LLL()
```
This only moves signs around.


mratsim:
Rebased and reopened in #405

Issue #357, opened 2024-02-11
Tracking compiler inefficiencies


mratsim:
As mentioned in https://github.com/mratsim/constantine/blob/661a481/README-PERFORMANCE.md#compiler-caveats
Compilers have a hard time optimizing bigint operations, even as simple as an addition with carries.

This issue track their evolution and the quality of the code generated with compiler builtins for ISAs of interest.

Note that as of February 2024, we use:
- `_addcarry_u64` on x86-64
- uint128 on other ISAs

## 2019, GCC 9.2 and Clang 9.0

The original problem:
https://gcc.godbolt.org/z/2h768y

![image](https://github.com/mratsim/constantine/assets/22738317/3073cea5-ff71-4d34-843a-86810dad26b1)

Even with intrinsics, an  operation as simple as addition-with-carry is uglily implemented in GCC.
This has been mentioned by the GMP folks 30 years ago: https://gmplib.org/manual/Assembly-Carry-Propagation.html

## 2024, GCC 13.2 and Clang 17.0

https://gcc.godbolt.org/z/jdecvffaP

![image](https://github.com/mratsim/constantine/assets/22738317/d3a2ff9d-5fbe-4366-9fbe-ab68a0cd694b)

GMP fixed the x86 intrinsics but unfortunately the portable intrinsics has a terribad codegen and hence makes a terrible fallback for ARM.

## Current status

Due to GCC abysmal __builtin__addcll it is a non-starter.
Clang has decent codegen.

Assembly is still very much needed.

This also explains the bad ARM performance on Apple M1, M2, M3, mentioned by @agnxsh (https://github.com/mratsim/constantine/pull/354#issuecomment-1925688995) and @bkomuves

mratsim:
According to https://stackoverflow.com/questions/33690791/producing-good-add-with-carry-code-from-clang#comment129749304_73472447, `__builtin__addcll` is useful on Clang since Clang 10 in 2020.

Fortunately, Clang 10 was released in March 2020, just prior to Ubuntu LTS 20.04, and its the default compiler there:
- https://packages.ubuntu.com/focal/clang
And Apple M1 or later which have the largest number of ARM users is from November 2020 so should also have proper support (unless Apple didn't merge all upstream improvements)

chfast:
The `__builint_addcll` has been added in GCC 14 because of the `_BitInt`. Your example of GCC 13 compiles only because C language allows undefined functions. It's good to enable some warnings at least `-Wall -Wextra`.

[...]

PR #381, opened 2024-05-09
C API for BLS signature `aggregate_verify` & `batch_verify` (incl parallel)

Vindaar:
Adds the C API for BLS signature `aggregate_verify` and `batch_verify` functions. The parallel version using a threadpool is also exposed.

Note: For the batched API we currently provide a `ByteView` typedef on the C side (happy to choose a different name, probably better snake_case and lower case, maybe `ctt_byte_view` ?). We _could_ do an anonymous struct in the function signatures, but at least I'm not aware of a way to get around 'incompatible pointer' warnings with GCC / Clang in that case.

It includes 2 minor fixes for the `batch_verify(_parallel)` Nim C targeted procs. They were accidentally generic and missing a `toOpenArray` for the `messages`. In addition it fixes a minor bug in the `update` proc for the `View[byte]` overload, which lead to an infinite recursion.

### DIFF
--- a/examples-c/README.md
+++ b/examples-c/README.md
@@ -8,7 +8,7 @@ in `lib/`.
 To compile and run the example / test case, for example:
 
 ```sh
-gcc ethereum_bls_signatures.c -o ethereum_bls_signatures -I../include -L../lib -lconstantine
+clang ethereum_bls_signatures.c -o ethereum_bls_signatures -I../include -L../lib -lconstantine
 ```
 
 For the test case, you also need to link in `-lgmp`.
@@ -17,7 +17,7 @@ And depending on your setup you might need to specify where
 `libconstantine.so` can be found:
 
 ```sh
-LD_LIBRARY_PATH=../lib ./ethereum_bls_signutares
+LD_LIBRARY_PATH=../lib ./ethereum_bls_signatures
 ```
 
 (in case you compile and run from this directory).
--- a/examples-c/README.md
+++ b/examples-c/README.md
@@ -2,4 +2,26 @@
 
 This folder holds tests (prefixed with `t_`) and examples for the C bindings.
 
-Headers are located in `include/` and the static and dynamic libraries in `lib/`
+Headers are located in `include/` and the static and dynamic libraries
+in `lib/`.
+
+To compile and run the example / test case, for example:
+
+```sh
+gcc ethereum_bls_signatures.c -o ethereum_bls_signatures -I../include -L../lib -lconstantine
+```
+
+For the test case, you also need to link in `-lgmp`.
+
+And depending on your setup you might need to specify where
+`libconstantine.so` can be found:
+
+```sh
+LD_LIBRARY_PATH=../lib ./ethereum_bls_signutares
+```
+
+(in case you compile and run from this directory).
+
+The above of course assumes you have already compiled
+`libconstantine.so` (using `nimble make_lib` from the root directory
+of the repository).
--- a/examples-c/ethereum_bls_signatures.c
+++ b/examples-c/ethereum_bls_signatures.c
@@ -64,7 +64,14 @@ int main(){
       { message, 32 }
   };
   const ctt_eth_bls_signature sigs[3] = { sig, sig, sig };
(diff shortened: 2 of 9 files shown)


examples-c/README.md
```diff
@@ -2,4 +2,26 @@
...
 
-Headers are located in `include/` and the static and dynamic libraries in `lib/`
\ No newline at end of file
+Headers are located in `include/` and the static and dynamic libraries
+in `lib/`.
+
+To compile and run the example / test case, for example:
+
+```sh
+gcc ethereum_bls_signatures.c -o ethereum_bls_signatures -I../include -L../lib -lconstantine
```
mratsim, line 11:
```suggestion
clang ethereum_bls_signatures.c -o ethereum_bls_signatures -I../include -L../lib -lconstantine
```

I'd rather use Clang everywhere as example.

GCC performance is just bad https://github.com/mratsim/constantine/issues/357

examples-c/README.md
```diff
@@ -2,4 +2,26 @@
...
+```
+
+For the test case, you also need to link in `-lgmp`.
+
+And depending on your setup you might need to specify where
+`libconstantine.so` can be found:
+
+```sh
+LD_LIBRARY_PATH=../lib ./ethereum_bls_signutares
```
mratsim, line 20:
```suggestion
LD_LIBRARY_PATH=../lib ./ethereum_bls_signatures
```

include/constantine/protocols/ethereum_bls_signatures.h
```diff
@@ -50,6 +50,13 @@ static const char* ctt_eth_bls_status_to_string(ctt_eth_bls_status status) {
   return "cttEthBls_InvalidStatusCode";
 }
 
+// Wrapper of the View[T] Nim type for the common case of View[byte]
+//
+// type View*[byte] = object # with T = byte
+//  data: ptr UncheckedArray[byte] # 8 bytes
+//  len*: int                      # 8 bytes (Nim `int` is a 64bit int type)
+typedef struct { byte* data; size_t len; } ByteView;
```
mratsim, line 58:
```suggestion
typedef struct { byte* data; size_t len; } ctt_span;
```

C++20 and C# have settled to call this abstraction a span, see https://www.cppstories.com/2023/span-cpp20/, and original proposal: https://open-std.org/JTC1/SC22/WG21/docs/papers/2016/p0122r1.pdf

I think we should follow suit.

Vindaar: If we 'embrace' it like this, I guess we could consider moving it to `core/datatypes.h` as well?

mratsim: Sounds good to my. Yes we can call it `ctt_span_bytes`, but I think it's fine for now, it's only used for arrays of messages. For the others we just have pointer+length.

examples-c/ethereum_bls_signatures.c
```diff
@@ -49,7 +49,49 @@ int main(){
...
+
+  // try to use batch verify; We just reuse the data from above 3 times
+  const ctt_eth_bls_pubkey pkeys[3] = { pubkey, pubkey, pubkey };
+  ctt_span messages[3] = { // already hashed message, reuse 3 times
+      { message, 32 },
+      { message, 32 },
+      { message, 32 }
+  };
+  const ctt_eth_bls_signature sigs[3] = { sig, sig, sig };
+  byte srb[32] = {0}; // just a bunch of zeros as random "secure" bytes.
```
mratsim, line 67: a bunch of 7, 8 or anything is better, because internally we multiply the randomness with the inputs and multiplying with 0 gives 0, and pairing of a zero input always gives 1.

To prevent that we re-hash here
https://github.com/mratsim/constantine/blob/2a2bf9c0c777d283265ce66632b43c9a0f9e1cb7/constantine/signatures/bls_signatures.nim#L404-L411

But I think it is a good occasion to show sysrand

mratsim (dismissed):
A not on the randomness source, otherwise LGTM!

mratsim (approved):

PR #382, opened 2024-05-09
C API for EVM precompiles

Vindaar:
This is based on and continues from #381. It now also adds a C API for the EVM precompiles module.

One example is added that generally checks the API works and it includes a test case from:

https://github.com/mratsim/constantine/blob/master/tests/protocol_ethereum_evm_precompiles/eip-2537/map_fp2_to_G2_bls.json

which passes as expected.

PR #383, opened 2024-05-14
Go API for BLS signatures

Vindaar:
Still a lot of cleanup to do. But wanted to push it at least.

Aside from cleanup, I need to do a careful pass of the code again to see if I didn't mess up anything. All the tests pass though, so that's a good start (but who knows, maybe I accidentally pass some tests somewhere...).

edit:

Cleaned it up a bit. Still need to add the parallel batch verify + test case.

Still, it's now at a reasonable point. The main questions I still have:

- Should the BLS signature types rather be

```go
type EthBlsSecKey struct {
	cSec C.ctt_eth_bls_seckey
}
```
  or
```go
type EthBlsSecKey C.ctt_eth_bls_seckey
```
  ?
  The latter would allow to just pass the memory of the Go slice of
  `[]EthBlsFoo` types to C.
- Naming of the Go wrapped functions: Remove or keep redundant
  information?
  i.e. `PubKeyIsZero` or just `IsZero`? `SerializePubKeyCompressed` or
  just `SerializedCompressed`?
  Given that the functions are bound to the types, the short ones
  should be fine? Although I saw that even the Nim API uses the 'long'
  names, e.g. `deserialize_pubkey_compressed_unchecked`.


mratsim:
Still, it's now at a reasonable point. The main questions I still have:

Should the BLS signature types rather be

> type EthBlsSecKey struct { 	cSec C.ctt_eth_bls_seckey }

or

> type EthBlsSecKey C.ctt_eth_bls_seckey

The second one, with type alias. There is no special init procedure required since they are all stack types so that's fine.

And short IsZero is fine.

The C API has long name because C has no namespacing so conflicts are easy when dealing with generic maths.

Vindaar:
> The second one, with type alias. There is no special init procedure required since they are all stack types so that's fine.

Got it, good!
 
> And short IsZero is fine.
> The C API has long name because C has no namespacing so conflicts are easy when dealing with generic maths.

I assumed the fact that Nim also uses some verbose names meant it was done for a reason. Will shorten them!

Vindaar:
Rebased onto the EVM precompiles PR (will rebase onto master, once #382 is merged), updated the naming to be short hand everywhere and added the parallel batch verify API.

constantine-go/constantine.go
```diff
@@ -265,4 +266,377 @@ func (ctx EthKzgContext) VerifyBlobKzgProofBatchParallel(tp Threadpool, blobs []
...
+	}
+	return spans
+}
+
+func freeSpans(spans []CttSpan){
+	// frees the malloc'd memory of the given spans
+	for _, msg := range spans {
+		C.free(unsafe.Pointer(msg.data))
+	}
+}
```
Vindaar, line 558: If we're happy with the copy approach for now, I'd go ahead and remove the not working `toSpan` here.

bindings/lib_constantine.nim
```diff
@@ -26,5 +26,7 @@ import
   ../constantine/ethereum_eip4844_kzg,
   ../constantine/ethereum_eip4844_kzg_parallel,
 
+  ../constantine/ethereum_evm_precompiles,
```
mratsim, line 29: seems like a rebase is needed after #382

constantine-go/constantine.go
```diff
@@ -265,4 +266,377 @@ func (ctx EthKzgContext) VerifyBlobKzgProofBatchParallel(tp Threadpool, blobs []
...
+
+func (pub *EthBlsPubKey) DerivePubKey(sec EthBlsSecKey) (bool, error) {
+	C.ctt_eth_bls_derive_pubkey((*C.ctt_eth_bls_pubkey)(pub), (*C.ctt_eth_bls_seckey)(&sec))
+	return true, nil
+}
+
+func (pub *EthBlsPubKey) Verify(message []byte, sig EthBlsSignature) (bool, error) {
+	status := C.ctt_eth_bls_verify((*C.ctt_eth_bls_pubkey)(pub),
+		(*C.byte)(unsafe.Pointer(&message[0])),
+		(C.ptrdiff_t)(len(message)),
```
mratsim, line 346: To schedule for later, ptr+len should use ptr+size_t as per C convention, and we cast to Nim int.

constantine-go/constantine.go
```diff
@@ -265,4 +266,377 @@ func (ctx EthKzgContext) VerifyBlobKzgProofBatchParallel(tp Threadpool, blobs []
...
+// Go pointer we pass to C, which points to Go pointers.
+func toSpan(data *[]byte) (span C.ctt_span) { //CttSpan) {
+	span = C.ctt_span{data: (*C.byte)(unsafe.Pointer(&(*data)[0])), len: C.size_t(len(*data))} // in place span
+	return span
+}
+
+func newSpans(data [][]byte) (spans []CttSpan) {
+	// TODO: this is ugly, but due to Go's rules about not allowing to
+	// pass a Go pointer, which itself contains a Go pointer, we cannot
+	// pass the data of each `message` to C... So gotta malloc and copy
```
mratsim, line 543: To avoid the newSpans / freeSpans, it would probably be best in another PR to expose the BlsBatchSigAccumulator: https://github.com/mratsim/constantine/blob/285aeb280f84b889195529bf821e2801423bcb08/constantine/signatures/bls_signatures.nim#L265-L313

And reimplement aggregate_verify and batch_verify of Go using that so there is no allocations and no copies:

https://github.com/mratsim/constantine/blob/285aeb280f84b889195529bf821e2801423bcb08/constantine/signatures/bls_signatures.nim#L582-L656

mratsim (approved):

Vindaar:
Rebased onto current master and fixed a part of one test case (was checking for `err != nil` after calling a function that can never return not nil).

The fix made me think about something I had forgotten to ask: I just used the `(bool, error)` return type for the entire API to follow the existing code and have a unified API. However, in the BLS signature Nim / C API there are some functions that don't have a "fail state" (signature is zero, pubkeys/signatures are equal). Should we just return `bool` in those cases?

mratsim:
> The fix made me think about something I had forgotten to ask: I just used the `(bool, error)` return type for the entire API to follow the existing code and have a unified API. However, in the BLS signature Nim / C API there are some functions that don't have a "fail state" (signature is zero, pubkeys/signatures are equal). Should we just return `bool` in those cases?

Yes

constantine-go/constantine.go
```diff
@@ -265,4 +266,428 @@ func (ctx EthKzgContext) VerifyBlobKzgProofBatchParallel(tp Threadpool, blobs []
...
+	}
+	return spans
+}
+
+func freeSpans(spans []CttSpan){
+	// frees the malloc'd memory of the given spans
+	for _, msg := range spans {
+		C.free(unsafe.Pointer(msg.data))
+	}
+}
```
mratsim, line 515: I think we can delete the Spans now?

constantine-go/constantine.go
```diff
@@ -265,4 +266,428 @@ func (ctx EthKzgContext) VerifyBlobKzgProofBatchParallel(tp Threadpool, blobs []
...
+		)
+		return false, err
+	}
+
+	return true, nil
+}
+
+// NOTE: C.ctt_eth_bls_batch_sig_accumulator is an incomplete struct. Therefore
+// we use 2 functions on the Nim side to (de)allocate storage for the struct.
+type EthBlsBatchSigAccumulator struct {
```
mratsim, line 603: The struct and its method can be private (lowercase) for now.

constantine-go/constantine.go
```diff
@@ -265,4 +266,428 @@ func (ctx EthKzgContext) VerifyBlobKzgProofBatchParallel(tp Threadpool, blobs []
...
+}
+
+func (accum EthBlsBatchSigAccumulator) FinalVerify() bool {
+	status := C.ctt_eth_bls_final_verify_batch_sig_accumulator(
+		(*C.ctt_eth_bls_batch_sig_accumulator)(accum.ctx),
+	)
+	return bool(status)
+}
+
+func BatchVerifyGo(pubkeys []EthBlsPubKey, messages [][]byte, signatures []EthBlsSignature, secureRandomBytes [32]byte) (bool, error) {
```
mratsim, line 638: And we can rename this to BatchVerifySoA, and we likely want a BatchVerifyAoS with an arrays of triplets (pubkey, message, signature).

And we can reimplement AggregateVerify in terms of the BLS Sig accumulator.

The parallel BatchVerify is not a priority and can be done later, probably using goroutines as it doesn't require advanced multithreading capabilities, see impl https://github.com/mratsim/constantine/blob/285aeb280f84b889195529bf821e2801423bcb08/constantine/signatures/bls_signatures_parallel.nim#L71-L181

The Go API is important for fuzzing in go-ethereum/geth but a parallel implementation is not needed for fuzzing.

constantine/ethereum_bls_signatures.nim
```diff
@@ -510,3 +513,62 @@ func batch_verify*[Msg](pubkeys: openArray[PublicKey], messages: openarray[Msg],
   if verified:
     return cttEthBls_Success
   return cttEthBls_VerificationFailure
+
+func alloc_batch_sig_accumulator*(): ptr BatchSigAccumulator {.libPrefix: prefix_ffi.} =
+  ## Allocates using `alloc0` for a `BatchSigAccumulator`
+  ## so that it can remain an incomplete struct in the C header.
+  result = cast[ptr BatchSigAccumulator](alloc0(sizeof(BatchSigAccumulator)))
```
mratsim, line 520: Use `allocHeapAligned`. https://github.com/mratsim/constantine/blob/285aeb280f84b889195529bf821e2801423bcb08/constantine/platforms/allocs.nim#L101-L107

We don't want to use Nim allocator in Constantine for low-level primitives so that:
- We can easily debug / track allocations with regular tools like valgrind
- Memory is deallocated and not retained for caching
- We don't bring the Nim runtime into other languages, this means we don't need to call `NimMain` before using constantine.dll / constantine.so and it's a nice UX improvement.
- Allocator can be swapped, Nim's is tuned for real-time, but Constantine is used in 24/7 applications and in that case might benefit from jemalloc for the batch accumulator.

constantine/ethereum_bls_signatures.nim
```diff
@@ -510,3 +513,62 @@ func batch_verify*[Msg](pubkeys: openArray[PublicKey], messages: openarray[Msg],
...
+  ## verify individually but would verify while aggregated
+  ## https://ethresear.ch/t/fast-verification-of-multiple-bls-signatures/5407/14
+  ##
+  ## An optional accumulator separation tag can be added
+  ## so that from a single source of randomness
+  ## each accumulatpr is seeded with a different state.
+  ## This is useful in multithreaded context.
+
+  # TODO: The domain separation tag and accumSepTag are both hardcoded in constantine
+  # Just drop them in the arguments here and fill here?
```
mratsim, line 547: Technically they are hardcoded for _Ethereum_.

But if we implement the parallel batch verification, the accumSepTag will differ.
I think it's better to make that dynamic so we don't break APIs later on.

It's OK to hardcode the domain separation tag as the C name is `ctt_eth_bls_batch_sig_accumulator` and the Nim code is contained in `ehtereum_bls_signatures.nim`

mratsim (approved):
LGTM in general, some small fixes regarding the BlsBatchSigAccumulator to:
- use regular malloc/free
- not expose it publicly in go
- prepare for BatchVerifyParallel (which is not a priority)
- support both SoA and AoS batching without realloc

constantine-go/constantine.go
Vindaar: Renamed to `BatchVerifySoA` and added a `BatchVerifyAoS` with a `[]BatchVerifyTriplet` type as the argument.

Not quite sure how to reimplement `AggregateVerify` using the batch sig accumulator, because of the secure random bytes / my naive assumption of just reusing the aggregate signature for each (pkey/message) pair not actually giving the right result.

Vindaar: For future reference: We'll expose the aggregate accumulator on the Nim side similar to the batch sig accumulator at a later date.

mratsim (approved):

constantine/ethereum_bls_signatures.nim
```diff
@@ -510,3 +513,57 @@ func batch_verify*[Msg](pubkeys: openArray[PublicKey], messages: openarray[Msg],
   if verified:
     return cttEthBls_Success
   return cttEthBls_VerificationFailure
+
+func alloc_batch_sig_accumulator*(): ptr BatchSigAccumulator {.libPrefix: prefix_ffi.} =
+  ## Allocates using `alloc0` for a `BatchSigAccumulator`
+  ## so that it can remain an incomplete struct in the C header.
+  result = cast[ptr BatchSigAccumulator](allocHeapAligned(BatchSigAccumulator, alignment = 64))
```
mratsim, line 520: The casting is not needed by the way https://github.com/mratsim/constantine/blob/285aeb280f84b889195529bf821e2801423bcb08/constantine/platforms/allocs.nim#L101-L107

```suggestion
  result = allocHeapAligned(BatchSigAccumulator, alignment = 64)
```

mratsim (approved):

constantine/ethereum_bls_signatures.nim
Vindaar: Heh, good catch!

PR #387, opened 2024-05-30
Rust API for BLS signatures and EVM precompiles

Vindaar:
And here we are finally for the first part of the Rust API, namely BLS signatures.

This is still a draft, because there's again a few questions about how this should ideally be wrapped. Currently the code simply uses type aliases similar to the Go API. This is "fine", but leads to the user having to write code like:

```rust
        let mut pub_key = MaybeUninit::<EthBlsPubKey>::uninit();
        let status = deserialize_pubkey_compressed(pub_key.as_mut_ptr(), inp.as_ref());
        let pub_key = unsafe { pub_key.assume_init() };
```

which imo is rather ugly.

2 alternatives come to mind:
- at least for these types simply have `deserialize_*_compressed` (and friends) return the deserialized type directly. That way we can hide the `unsafe` code in the fake constructor. 
- wrap the C types in a Rust struct and abstract it away that way. This approach will just mean we have to do more work on API implementation side, but at least we get shorter function names (because we can then "overload" names like `deserialize`) and we shouldn't suffer from copies, because we can of course just pass the pointers of the underlying object to C.

Another point:
Funnily (or ironically), while we can nicely turn the nested messages in `batch_verify` etc. into a bunch of `CttSpan`s, we *still* have to make a copy of the data due to Rusts extreme stance on pointer mutability. Given that the `messages` argument is just a reference (wouldn't make sense to force it to be `&mut`), we are not allowed to turn them into a `CttSpan` (or `ctt_span` to be exact), because it is defined as having a `byte*` field and not a `const byte*` field. I'm a bit perplexed that we cannot even cast an immutable to a mutable pointer inside of an `unsafe` block. At least that's my understanding so far.
I'm not sure if we want to restrict us so far as to have the `ctt_span` have a `const byte*` field, because there should be valid reasons to use the type for mutation as well? We could of course have a `ctt_span_const`, but that's also a bit ridiculous.

Anyway, while the tests are not pretty due to so much error checking and `MaybeUninit` work, at least they all pass.


Vindaar:
Just finished porting over the EVM precompiles as well. Decided to just do it in the same PR now.

mratsim (approved):

PR #388, opened 2024-06-06
R1CS parser continuation

Vindaar:
This PR continues on from #327.

I tried to stick very closely to the binary layout of the files in the definition of the R1CS object. We could restructure it to be such that it has different types for different sections explicitly. But given the description of the file format, this is more honest. 

Still have to:
- add test cases
- decide if some types should be uint32/64 or int32/64
- change the type of `prime: seq[byte]` to something more appropriate. I assume I'll do the same as for the `value` of the constraints?


constantine/proof_systems/constraint_systems/r1cs_circom_parser.nim
```diff
@@ -20,15 +20,20 @@ import
...
+
 type
-  R1csSectionKind = enum
-    kHeader                 = 1
-    kConstraints            = 2
-    kWire2LabelId           = 3
-    kCustomGatesList        = 4
-    kGustomGatesApplication = 5
+  R1csSectionKind = enum        # Actual values according to spec in comments. To use it as discriminator in  variant object
+    kHeader                 = 0 # 1
```
mratsim, line 29: Is it necessary to not use the spec value? Otherwise we can create a kPlaceholder for value 0 if we want an enum without hole.

constantine/proof_systems/constraint_systems/r1cs_circom_parser.nim
```diff
@@ -40,22 +45,82 @@ type
...
+     ┏━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+     ┃ 64 │   01 00 00 00 00 00 00 00   ┃   nLabels
+     ┗━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
+     ┏━━━━┳━━━━━━━━━━━━━━━━━┓
+     ┃ 32 │   01 00 00 00   ┃               mConstraints
+     ┗━━━━┻━━━━━━━━━━━━━━━━━┛
+  ]#
+  Header = object
+    fieldSize: uint32 # field size in bytes (fs)
+    prime: seq[byte] # XXX: What type to use with RT size info?
```
mratsim, line 77: `seq[byte]` is fine, it would be bigEndian encoded prime number which is something often used in specs.

It is often called "octet string" in cryptographic spec:
- https://crypto.stackexchange.com/questions/37537/what-are-i2osp-os2ip-in-rsa-pkcs1
- https://www.rfc-editor.org/rfc/rfc3447#section-4.1

>    For the purposes of this document, and consistent with ASN.1 syntax,
   an octet string is an ordered sequence of octets (eight-bit bytes).
   The sequence is indexed from first (conventionally, leftmost) to last
   (rightmost).  For purposes of conversion to and from integers, the
   first octet is considered the most significant in the following
   conversion primitives.

constantine/proof_systems/constraint_systems/r1cs_circom_parser.nim
```diff
@@ -83,23 +149,117 @@ proc parseConstraint(f: File, constraint: var Constraint, fieldSize: int32): boo
...
+  result = true # would have returned before due to `?` otherwise
+
+proc parseWire2Label(f: File, v: var Wire2Label, sectionSize: uint64): bool =
+  r1csSection(sectionSize):
+    let numWires = sectionSize div 8
+    v.wireIds.setLen(numWires)
+    for labelId in v.wireIds.mitems():
       ?f.parseInt(labelId, littleEndian)
+
+## TODO: to be written :)
```
mratsim, line 195: Note that I am unaware of any usage of these so they are not in scope for this PR

constantine/proof_systems/constraint_systems/r1cs_circom_parser.nim
Vindaar: We can indeed just add a placeholder.

Vindaar:
I've added the test case from 
https://github.com/iden3/r1csfile/blob/master/test/r1csfile.js
using the file from here:
https://github.com/iden3/r1csfile/tree/master/test/testutils

tests/proof_systems/t_r1cs_parser.nim
```diff
@@ -0,0 +1,83 @@
...
+suite "R1CS binary file parser":
+
+  test "Parse basic example R1CS file":
+    # Note: The example test file used here is from:
+    # https://github.com/iden3/r1csfile/tree/master/test/testutils
+    const path = TestDir / "r1cs_test_files/example.r1cs"
+    let r1cs = parseR1csFile(path)
+
+    ## XXX: On 32bit systems `seq[BaseType]` should have twice as many elements
+    ## in the `expABCs` below as `value` arguments!
```
mratsim, line 16: Given that when converting to `Factor` we aren't even validating the elements, see:
- https://github.com/mratsim/constantine/pull/388/files#diff-f275b0ff9384fffa4fb5adc6a027fc7deaf68890029efd3a59a0911c047ee3adR36
- https://github.com/mratsim/constantine/pull/388/files#diff-f275b0ff9384fffa4fb5adc6a027fc7deaf68890029efd3a59a0911c047ee3adR128

We might as well have Factor use a `seq[byte]` to be uint32/uint64 agnostic. And do the conversion in a later pass, along with validation.

mratsim (approved):

Issue #367, opened 2024-03-09
IPA / Verkle tree cleanups


mratsim:
From https://github.com/mratsim/constantine/pull/364#pullrequestreview-1926092368

## Easy
- [x] there is a `newSeq` introduced that needs to be replaced to avoid dependency on Nim runtime, especially if we want to expose IPA through C/Go/Rust
- [x] some proc like absIntChecker are exported but might be local-only
- [x] scalarMul needs an overload for field element and not just big ints
- [x] views need a splitMiddle functions and StridedView can then be replaced

## Medium
- [x] Fixing verkle tests on Windows (out of stack space) #330
- [x] `createIPAProof` initialization is incomplete and need zero-init: https://github.com/mratsim/constantine/pull/389/commits/da069492243ad5763395840897fab6b3e3069802

## Complex

- [x] evalInDomain / evalOutside domain and the Lagrange form of IPA are duplicates of KZG the polynomial functions (used in EIP-4844) need to be deduplicated and parallelized. See #361.
- [ ] Constant-time dlog square root #358 and #359

mratsim:
Closed by #397

PR #389, opened 2024-06-07
PR #330 - enable test verkle ipa

mratsim:
This is #330, rebased to latest master and for investigation of CI failure on Windows


mratsim:
Bisecting, the regression is due to #371

mratsim:
Confirmed that Windows failure without any error is due to a stack overflow due to `array[VerkleDomain, array[VerkleDomain, Fr[Banderwagon]]` being too large for Windows stack.

The temporary workaround is to use a ref array, see https://github.com/mratsim/constantine/pull/389/commits/bc7490281279e68c112b5bcaddde27e7b6d56009.

Ultimately those should be replaced by using Constantine allocator functions, to avoid Nim allocator, see: https://github.com/mratsim/constantine/issues/367

Strangely, if changing to a ref alloc in the test file, the Multiproof creation and verification test fails. Hopefully switching 1 out of 2 to heap space leaves enough stack space for Windows test.

PR #393, opened 2024-06-11
reenable nim v2.0.x in CI

mratsim:
- Disabled in https://github.com/mratsim/constantine/pull/375
- Regression reverted in backport in https://github.com/nim-lang/Nim/commit/767a901267c782765e010d106fae277b27e04981


mratsim:
Latest bug regression should have been fixed https://github.com/nim-lang/Nim/issues/23711

PR #394, opened 2024-06-12
API: argument order (breaking external Nim/C/Go/Rust EIP-4844 API

mratsim:
This changes the threadpool to always be first argument. It is the most "global" of the context, because it's at library/app level and not just KZG or BLS signatures.

As powers-of-tau and polynomial domains are also contexts, they are moved in front and this impacts KZG and IPA modules.

Furthermore, document the API.

It is inspired by libsecp256k1:
- https://github.com/bitcoin-core/secp256k1/blob/v0.5.0/include/secp256k1.h#L10-L25

This breaks Nim, C, Go, Rust API and should be done now before we cut the first ever Constantine release.
This impacts Grandine's branch: https://github.com/mratsim/constantine/pull/331

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,35 @@
+# Architecture
+
+## APIs
+
+### Argument orders
+
+Function calls have arguments ordered the following way:
+
+1. Context arguments
+2. Threadpool context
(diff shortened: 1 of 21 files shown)


mratsim:
1 windows CI failed due to a bad mingw mirror
image

PR #411, opened 2024-06-28
refactor(add-carry): with Clang on non-x86 (for example MacOS) use builtin add-carry instead of u128

mratsim:
Potentially a quick win to accelerate Constantine on Apple M1~3.

See tracking compiler inefficiencies #357 

Use addcarry builtin for Clang (GCC is incredibad) on non-x86 platforms.

To be benched with `CTT_ASM=0 CC=clang nimble bench_fp`.


mratsim:
On a Apple M4 Max, this is a 10% perf improvement for scalar multiplication.
image

Issue #396, opened 2024-06-22
Verkle IPA - Incomplete test suite


mratsim:
The test suite for Verkle Tries IPA is incomplete.

We're missing at least:
- Transcript test: https://github.com/crate-crypto/verkle-trie-ref/blob/2332ab8/ipa/ipa_test.py#L38-L50
- Multiproof: https://github.com/crate-crypto/verkle-trie-ref/blob/2332ab8/multiproof/multiproof_test.py
- CRS: https://github.com/crate-crypto/go-ipa/blob/b1e8a79/ipa/ipa_test.go#L253-L264
- Multiproof: https://github.com/crate-crypto/go-ipa/blob/b1e8a79/multiproof_test.go#L63-L132
- Multiproof: https://github.com/crate-crypto/rust-verkle/blob/442174e/ipa-multipoint/src/multiproof.rs#L435-L509

And all the test vectors implemented here: https://github.com/jsign/verkle-test-vectors

In particular test011 is supposed to to negative testing and verify that multiproofs fail for incorrect inputs: https://github.com/jsign/verkle-test-vectors/blob/735b7d6/crypto/clients/go-ipa/crypto_test.go#L320-L326
```go
	// Walk the [0, 255] range and verify the proof fails for every evaluation point but the correct one.
	for i := 0; i < 256; i++ {
		transcript := common.NewTranscript("multiproof")
		ok, err := multiproof.CheckMultiProof(transcript, config, &proof, []*banderwagon.Element{&commitment}, []*fr.Element{&evalResult}, []uint8{uint8(i)})
		require.NoError(t, err)
		require.True(t, ok == (i == data.TestData.ForgedEvaluationResult.EvaluationPoint))
	}
```
but we don't do that https://github.com/mratsim/constantine/blob/9fe5e495f7f83c9a550c8a89204e28332787d3cd/tests/t_ethereum_verkle_ipa_primitives.nim#L702-L710

And when checking other evaluation points the test seems to always return true?

mratsim:
And the first multiproof test is flawed, Fs is of length 256 while the rest is length 1:
https://github.com/mratsim/constantine/blob/9fe5e495f7f83c9a550c8a89204e28332787d3cd/tests/t_ethereum_verkle_ipa_primitives.nim#L641-L659

And they should be same length:
[...]

PR #412, opened 2024-06-29
fixed IPA consistency test, rm addressSanitizer message

agnxsh:
Addresses #396 ,
Fixes IPA consistency test, now the test returns the correct `prover_challenge` post `ipa_prove`, and then correctly verifies the same correctly as well.

For reference, https://github.com/crate-crypto/verkle-trie-ref/blob/2332ab8/ipa/ipa_test.py#L38-L50.

Also removed the `compile with address sanitizer ON` message, as that bug has been now fixed here #403


mratsim (approved):
Thanks!

PR #417, opened 2024-07-02
[Go] [Rust] update language bindings to return value API

Vindaar:
Previously the API for Go and Rust simply followed the C API (and by extension the Nim API from which the C API is generated) almost literally. That meant return types were reserved for error codes (or enums) and all 'interesting' data was passed via mutable arguments.

This changes the API to return values instead. This has the nice advantage on the Rust side that it sidesteps the `MaybeUninit` issue mentioned in #387.

Note: currently the CI does not pass, because of the bug fixed in #409. There might be other minor issues present, but due to "expected" test failures, I haven't dug into every current failure for Go/Rust.


constantine-go/constantine.go
```diff
@@ -357,135 +390,136 @@ func (sig1 EthBlsSignature) AreEqual(sig2 EthBlsSignature) bool {
...
+func (pub *EthBlsPubKey) SerializeCompressed() (dst Bytes48, err error) {
+	status := C.ctt_eth_bls_serialize_pubkey_compressed((*C.byte)(unsafe.Pointer(&dst)),
 		(*C.ctt_eth_bls_pubkey)(pub),
 	)
 	if status != C.cttCodecEcc_Success {
 		err := errors.New(
 			C.GoString(C.ctt_codec_ecc_status_to_string(status)),
 		)
-		return false, err
+		return dst, err
```
mratsim, line 447:
```suggestion
		return _, err
```
I think we can return a default value here, I don't remember what Nim does there.

constantine-go/constantine.go
```diff
@@ -357,135 +390,136 @@ func (sig1 EthBlsSignature) AreEqual(sig2 EthBlsSignature) bool {
...
+func (sig *EthBlsSignature) SerializeCompressed() (dst Bytes96, err error) {
+	status := C.ctt_eth_bls_serialize_signature_compressed((*C.byte)(unsafe.Pointer(&dst)),
 		(*C.ctt_eth_bls_signature)(sig),
 	)
 	if status != C.cttCodecEcc_Success {
 		err := errors.New(
 			C.GoString(C.ctt_codec_ecc_status_to_string(status)),
 		)
-		return false, err
+		return dst, err
```
mratsim, line 460: ditto

constantine-go/constantine.go
```diff
@@ -357,135 +390,136 @@ func (sig1 EthBlsSignature) AreEqual(sig2 EthBlsSignature) bool {
...
+	status := C.ctt_eth_bls_deserialize_pubkey_compressed_unchecked((*C.ctt_eth_bls_pubkey)(&pub),
 		(*C.byte)(&src[0]),
 	)
-	if status != C.cttCodecEcc_Success {
+	if status != C.cttCodecEcc_Success && status != C.cttCodecEcc_PointAtInfinity {
 		err := errors.New(
 			C.GoString(C.ctt_codec_ecc_status_to_string(status)),
 		)
-		return false, err
+		return pub, err
```
mratsim, line 473: ditto

constantine-go/constantine.go
```diff
@@ -357,135 +390,136 @@ func (sig1 EthBlsSignature) AreEqual(sig2 EthBlsSignature) bool {
...
+	status := C.ctt_eth_bls_deserialize_signature_compressed_unchecked((*C.ctt_eth_bls_signature)(&sig),
 		(*C.byte)(&src[0]),
 	)
-	if status != C.cttCodecEcc_Success {
+	if status != C.cttCodecEcc_Success && status != C.cttCodecEcc_PointAtInfinity {
 		err := errors.New(
 			C.GoString(C.ctt_codec_ecc_status_to_string(status)),
 		)
-		return false, err
+		return sig, err
```
mratsim, line 486: ditto

constantine-go/constantine.go
```diff
@@ -357,135 +390,136 @@ func (sig1 EthBlsSignature) AreEqual(sig2 EthBlsSignature) bool {
...
+func DeserializePubKeyCompressed(src Bytes48) (pub EthBlsPubKey, err error) {
+	status := C.ctt_eth_bls_deserialize_pubkey_compressed((*C.ctt_eth_bls_pubkey)(&pub),
 		(*C.byte)(&src[0]),
 	)
 	if status != C.cttCodecEcc_Success && status != C.cttCodecEcc_PointAtInfinity {
 		err := errors.New(
 			C.GoString(C.ctt_codec_ecc_status_to_string(status)),
 		)
-		return false, err
+		return pub, err
```
mratsim, line 499: ditto

constantine-go/constantine.go
```diff
@@ -357,135 +390,136 @@ func (sig1 EthBlsSignature) AreEqual(sig2 EthBlsSignature) bool {
...
+func DeserializeSignatureCompressed(src Bytes96) (sig EthBlsSignature, err error) {
+	status := C.ctt_eth_bls_deserialize_signature_compressed((*C.ctt_eth_bls_signature)(&sig),
 		(*C.byte)(&src[0]),
 	)
 	if status != C.cttCodecEcc_Success && status != C.cttCodecEcc_PointAtInfinity {
 		err := errors.New(
 			C.GoString(C.ctt_codec_ecc_status_to_string(status)),
 		)
-		return false, err
+		return sig, err
```
mratsim, line 512: ditto

constantine-go/constantine.go
mratsim: Or maybe dst is already default init. Anyway we should assume that people check the error.

constantine-go/constantine.go
```diff
@@ -657,212 +695,234 @@ func BatchVerifyAoS(triplets []BatchVerifyTriplet, secureRandomBytes [32]byte) (
...
+		return result, err
 	}
-	return true, nil
+	return result, nil
 }
 
-func EvmBls12381G1Add(result []byte, inputs []byte) (bool, error) {
-	status := C.ctt_eth_evm_bls12381_g1add((*C.byte)(getAddr(result)),
-		(C.ptrdiff_t)(len(result)),
+// TODO: continue from here
```
mratsim, line 793: to be removed

constantine-rust/constantine-ethereum-evm-precompiles/tests/t_ethereum_evm_precompiles.rs
```diff
@@ -89,13 +89,49 @@ fn from_hex(hex: HexString) -> Result<Vec<u8>, bool> {
...
     let mut result = vec![0u8; href.len() / 2];
     match hex::decode_to_slice(href, result.as_mut_slice() as &mut [u8]) {
-        Ok(_) => Ok(result),
-        Err(_) => Err(false),
+        Ok(_) => Some(result),
+        Err(_) => None,
+    }
+}
+
+trait CompareReturn {
```
mratsim, line 97: I think we can require `impl Iterator<Item = u8>`

constantine-rust/constantine-ethereum-evm-precompiles/tests/t_ethereum_evm_precompiles.rs
```diff
@@ -89,13 +89,49 @@ fn from_hex(hex: HexString) -> Result<Vec<u8>, bool> {
...
+}
 
-fn t_generate(test_name: String, func: TestFunction) {
+impl CompareReturn for Vec<u8> {
+    fn compare(self, v: Vec<u8>) -> bool {
+        return self == v;
+    }
+}
+
+fn t_generate<T: CompareReturn, F>(test_name: String, func: F)
```
mratsim, line 131:
```suggestion
fn t_generate<T: impl Iterator<Item = u8>, F>(test_name: String, func: F)
```

here

constantine/ethereum_evm_precompiles.nim
```diff
@@ -60,6 +60,31 @@ func eth_evm_sha256*(r: var openArray[byte], inputs: openArray[byte]): CttEVMSta
   sha256.hash(cast[ptr array[32, byte]](r[0].addr)[], inputs)
   return cttEVM_Success
 
+func modexp_result_size*(size: var uint64, inputs: openArray[byte]): CttEVMStatus {.noInline, tags:[Alloca, Vartime], libPrefix: prefix_ffi, meter.} =
```
mratsim, line 63:
```suggestion
func eth_evm_modexp_result_size*(size: var uint64, inputs: openArray[byte]): CttEVMStatus {.noInline, tags:[Alloca, Vartime], libPrefix: prefix_ffi, meter.} =
```

include/constantine/protocols/ethereum_evm_precompiles.h
```diff
@@ -56,10 +56,28 @@ static const char* ctt_evm_status_to_string(ctt_evm_status status) {
...
+ *  in the call to `eth_evm_modexp` based on this function's result.
+ *
+ *  The size depends on the `modulusLen`, which is the third 32 bytes,
+ *  `inputs == [baseLen { 32 bytes }, exponentLen { 32 bytes }, modulusLen { 32 bytes }, ... ]`
+ *  in `inputs`.
+ *
+ *  The associated modulus length in bytes is the size required by the
+ *  result to `eth_evm_modexp`.
+ */
+ctt_evm_status ctt_modexp_result_size(
```
mratsim, line 75:
```suggestion
ctt_evm_status ctt_eth_evm_modexp_result_size(
```

mratsim (approved):
Thanks, lgtm, some nits

constantine-rust/constantine-ethereum-evm-precompiles/src/lib.rs
```diff
@@ -9,259 +9,256 @@
...
+            _ => Err(status),
         }
     }
 }
 
-pub fn evm_bls12381_map_fp2_to_g2(
-    result: &mut [u8],
-    inputs: &[u8]
-) -> Result<bool, ctt_evm_status> {
+pub fn evm_bls12381_map_fp2_to_g2(inputs: &[u8]) -> Result<[u8; 256], ctt_evm_status> {
```
mratsim, line 250:
```suggestion
#[inline]
pub fn evm_bls12381_map_fp2_to_g2(inputs: &[u8]) -> Result<[u8; 256], ctt_evm_status> {
```

constantine/ethereum_evm_precompiles.nim
Vindaar: I wasn't sure whether to reuse the naming convention. I didn't so that there is a clearer distinction between what is a real EVM precompiles and what is a helper function. Will change it.

constantine-go/constantine.go
Vindaar: > I think we can return a default value here, I don't remember what Nim does there.

Returning `_` is not valid in Go. We can either do `TheType{}` to explicitly zero mem or just rely on the fact that all values we return are stack allocated and thus zeroed anyway. 

On the Nim side (in the actual functions that implement the logic, not the module that generates the C API) it seems like in some cases we use `setZero` e.g. here:

https://github.com/mratsim/constantine/blob/master/constantine/serialization/codecs_bls12_381.nim#L109-L111

and in other cases we simply don't touch the value we return, e.g. here:

https://github.com/mratsim/constantine/blob/master/constantine/serialization/codecs_bls12_381.nim#L148-L149

constantine-rust/constantine-ethereum-evm-precompiles/tests/t_ethereum_evm_precompiles.rs
Vindaar: While your proposal is not quite correct (`impl Iterator` is not valid syntax must be without the `impl`), it led me down the right path. `Iterator` would require the test functions to actually return an iterator. But there's also a predefined `IntoIterator` trait, which works (via `x.into_iter().collect::<Vec<u8>>()`).

constantine/ethereum_evm_precompiles.nim
mratsim: It's namespacing in case several protocols have differing modexp, say a modexp for the EVM, a modexp to simulate GMP, etc ...

constantine-rust/constantine-ethereum-evm-precompiles/tests/t_ethereum_evm_precompiles.rs
mratsim: Ah yes,  the impl Iterator is for arguments, not generics. And yes Vec use `into_iter` indeed.

mratsim (approved):
Awesome, thanks a lot

Issue #299, opened 2023-11-20
Polynomial Commitments


mratsim:
## Production-grade commitments

KZG is implemented and IPA is WIP for verkle trees (#275).

- FRI commitments https://eprint.iacr.org/2020/654

## Multilinear

The SNARKs world is moving to multilinear polynomial commitment schemes (multilinear PCS), in particular to remove the need of large FFTs that require a lot of memory, and scale with O(n log n).

### Commiting to a multilinear polynomial

These schemes can be composed on top of an univariate PCS like KZG10:

- Gemini: https://eprint.iacr.org/2022/420
- LogUp+: https://eprint.iacr.org/2023/1284
- Zeromorph: https://eprint.iacr.org/2023/917

### Multilinear PCS

- BCG20: https://eprint.iacr.org/2020/1426
- Brakedown, Ligero & Shockwave: https://eprint.iacr.org/2021/1043

mratsim:
A very promising one: Binius

> We introduce an efficient SNARK for towers of binary fields. Adapting Brakedown (CRYPTO '23), we construct a multilinear polynomial commitment scheme suitable for polynomials over tiny fields, including that with 2 elements. Our commitment scheme, unlike those of previous works, treats small-field polynomials with zero embedding overhead. We further introduce binary-field adaptations of HyperPlonk's (EUROCRYPT '23) product and permutation checks, as well as of Lasso's lookup. Our scheme's binary PLONKish variant captures standard hash functions—like Keccak-256 and Grøstl—extremely efficiently. With recourse to thorough performance benchmarks, we argue that our scheme can efficiently generate precisely those Keccak-256-proofs which critically underlie modern efforts to scale Ethereum.

https://eprint.iacr.org/2023/1784
https://gitlab.com/UlvetannaOSS/binius

PR #423, opened 2024-07-06
Multilinear extensions of polynomials

mratsim:
This implements multilinear extensions of polynomials.

This prepares for a focus on efficient proof systems and underlies:
- the sumcheck protocol (for linear-time polynomial IOP/Interactive Oracle Proof),
- the GKR protocol (for efficient circuits like for hash functions or machine learning)
- and multilinear polynomial commitments #299


constantine/boolean_hypercube/multilinear_extensions.nim
```diff
@@ -0,0 +1,318 @@
...
+    ## which takes value 1 at 𝛘ₑ(e) and 0 at Xₑ(e̅) e̅ any other element ∈ {0,1}ˢ
+    ##
+    ## 𝛘ₑ(x₁, ...,xₛ) = ∏ᵢ₌₁ˢ(xᵢeᵢ + (1-xᵢ)(1-eᵢ)), i ∈ [1, s]
+    ##
+    ## A multilinear polynomial is linear (i.e. degree at most 1) in each
+    ## of its variables.
+    num_vars*: int
+    base_poly_evals*: ptr UncheckedArray[F]
+
+proc `=destroy`*[F](mle: var MultilinearExtension[F]) =
```
Vindaar, line 43: If we attach a destructor to a plain (non ref) `object`, we should better error on `=copy`, no? Otherwise there is a risk of a copy of a `MultilinearExtension` object going out of scope and calling the destructor and thus freeing the memory. 

(my experience with `object` + `=destroy` has been less than splendid).

mratsim: Right, that or check if nil before calling free.

Vindaar: No, I mean not in terms of a potential double free, but in terms of having a dangling pointer in the instance that outlives the other. One might accidentally free the storage too early.

mratsim: Addressed in https://github.com/mratsim/constantine/pull/423/commits/acbfc482eb78fa762406c2a3f86e4fa17a5b112b

PR #424, opened 2024-07-07
add: multiproof consistency test

agnxsh:
Adds a Verkle multiproof consistency test, a continuation to addressing issue #396, implements the test https://github.com/crate-crypto/go-ipa/blob/b1e8a79/multiproof_test.go#L63-L132 as per the Verkle trie spec

Skips ipa_multi_verify because of a potential bug in `g2t`, will be addressed and fixed in successive PRs

### DIFF
--- a/tests/t_ethereum_verkle_ipa_primitives.nim
+++ b/tests/t_ethereum_verkle_ipa_primitives.nim
@@ -596,9 +596,81 @@ suite "IPA proof tests":
 # TODO: refactor completely the tests - https://github.com/mratsim/constantine/issues/396
 
 suite "Multiproof Tests":
-  echo "Warning! - Skipping all but serialization tests due to issues outlined in https://github.com/mratsim/constantine/issues/396"
+  echo "Warning! - Skipping all but serialization tests and Multiproof consistency test due to issues outlined in https://github.com/mratsim/constantine/issues/396"
   # The comparison between previous and new implementation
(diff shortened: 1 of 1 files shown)


mratsim (approved):
Very nice, so this confirms that generating multiproof is bug free.

agnxsh:
> Very nice, so this confirms that generating multiproof is bug free.

Absolutely 🙂

mratsim:
Currently investigating the g2t issue. I found another bug in the multiproof generation so we need more complete coverage: https://github.com/mratsim/constantine/commit/1f8c376a1f97a5082e54bce5930d48eabbbbf705

mratsim:
To clarify, the PR added the following test

https://github.com/mratsim/constantine/blob/618d8b3492980f12f7a51e7946abaca4469442f5/tests/t_ethereum_verkle_ipa_primitives.nim#L658-L667

to confirm that the state of the transcript matches with go-ipa.

However `getQuotientPolyInDomain` was wrongly implemented and was missing a multiplication here https://github.com/mratsim/constantine/commit/1f8c376a1f97a5082e54bce5930d48eabbbbf705

meaning g was wrong here: https://github.com/mratsim/constantine/blob/618d8b3492980f12f7a51e7946abaca4469442f5/constantine/commitments/eth_verkle_ipa.nim#L746-L769

D is derived from committing to g(X) and then absorbed into the transcript. I'm not sure why the transcript was correct for that consistency test but it does outline that we need a more comprehensive test suite #396

PR #429, opened 2024-07-14
𝔾ₜ exponentiation, with endomorphism acceleration

mratsim:
This PR adds a performance baseline for Whisk in GT.

https://ethresear.ch/t/the-return-of-torus-based-cryptography-whisk-and-curdleproof-in-the-target-group/16678

Here is the speed on a 7840U:

image

For comparison here is scalar mul on G1
image

and on G2
image

Hence exponentiation in Gt (131 us) is within 30% of scalar mul in G2 (96 us).

cc @asanso


mratsim:
Gnark bench

```
cd ecc/bls12-381
go test -bench=ExpGT -run=none --cpu 1
```

image

so 2.29x faster than Gnark.

PR #439, opened 2024-07-20
fix(nvidia): reorg + rename following #402

mratsim:


constantine/math/extension_fields/towers.nim
```diff
@@ -57,8 +57,8 @@ type
     CubicExt[Fp2[Name]]
 
   Fp12*[Name: static Algebra] =
-    CubicExt[Fp4[Name]]
-    # QuadraticExt[Fp6[Name]]
+    # CubicExt[Fp4[Name]]
+    QuadraticExt[Fp6[Name]]
```
mratsim, line 61: committed by mistake

Issue #285, opened 2023-10-17
[Code golfing] secp256k1 programming language benchmark


mratsim:
Following https://forum.nim-lang.org/t/10550#70556, this outlines step to put Nim at the top of https://programming-language-benchmarks.vercel.app/problem/secp256k1

- [ ] **Benchmark the top implementations.**
    Rust and Go on your machine. Go is using the standard libsecp256k1 library, which is alsowrapped in Nim: https://github.com/status-im/nim-secp256k1
- [ ] **Implement addition chain for 21**
    Here: https://github.com/mratsim/constantine/blob/34baa74bc0eb2504a939df35c6e6fcea761cc52a/constantine/math/arithmetic/finite_fields.nim#L443-L456
    This is blocking for projective coordinates benchmarking due to this: https://github.com/mratsim/constantine/blob/34baa74bc0eb2504a939df35c6e6fcea761cc52a/constantine/math/elliptic/ec_shortweierstrass_projective.nim#L231
    For secp256k1, curve equation y² = x³ + ax + b, with a=0 and b=7.
  - Optimal shortest addition chains can be found here: https://wwwhomes.uni-bielefeld.de/achim/addition_chain.html
- [ ] **Benchmark Constantine on the same computer to have a baseline.**
    Projective coordinates will need to be commented out if previous step has been skipped, endomorphism acceleration will need to be commented out.
- [ ] **Add endomorphism acceleration for secp256k1.**
    The constants can be generated from https://github.com/mratsim/constantine/blob/master/sage/derive_endomorphisms.sage. Note there is no G1 or G2 for non-pairing-friendly curves like secp256k1 or Banderwagon so the Sage code needs to be modified to handle both case. The constants can then be added to https://github.com/mratsim/constantine/tree/master/constantine/math/constants.
  - Difficulty: Easy, mostly existing Python/sage code modification, for generating Nim code. The scaffolding exist.
  - Expected speedup: 30% compared to baseline
  - Reference paper: https://link.springer.com/content/pdf/10.1007/3-540-44647-8_11.pdf
- [ ] **Add #73** 
  - Difficulty: Very hard
  - Expected speedup: 80% compared to endomorphism, 2.34x compared to baseline.
  - Reference paper:
    Efficient and Secure Algorithms for GLV-Based Scalar \
    Multiplication and their Implementation on GLV-GLS  \
    Curves (Extended Version) \
    Armando Faz-Hernández, Patrick Longa, Ana H. Sánchez, 2013 \
[...]

PR #444, opened 2024-07-26
feat(secp256k1): add endomorphism acceleration

mratsim:
This is the first of a series of PR to bring secp256k1 performance to state-of-the-art and then beyond.

See overall steps in #285.

## Benchmarks

Unfortunately the most important primitive field multiplication does not use assembly for secp256k1 (256-bit) at the moment. Hence the huge gap (~30%) with the Pallas curve (255-bit) for example.

image
image

image
image

### DIFF
--- a/benchmarks/bench_ec_g1.nim
+++ b/benchmarks/bench_ec_g1.nim
@@ -34,7 +34,7 @@ const AvailableCurves = [
   BN254_Snarks,
   # Edwards25519,
   # P256,
-  # Secp256k1,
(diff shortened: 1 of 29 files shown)


mratsim:
Bench from 

image
https://github.com/RustCrypto/elliptic-curves/blob/72540f8/k256/benches/scalar.rs#L36-L43

35us vs our 46us so we need a 24% perf improvement to catchup. This should be easy by not using Montgomery domain and implementing fast reduction.

PR #466, opened 2024-09-10
Implement finite field `ccopy`, `neg`, `cneg`, `nsqr`, ... for CUDA target

Vindaar:
This extends the existing `add`, `mul`, `sub` operations on finite fields on the Nvidia target by `ccopy`, `neg`, `cneg`, `nsqr`. For `nsqr` we have two different implementations for now. 
1. generate N squaring operations based on a Nim runtime / JIT compile time value 
2. generate a loop on the GPU using a phi node and branches for JIT runtime values

I added a simple test case for each operation, which compares with the equivalent operation on the CPU. Currently I just cloned the test file for each case. It'd probably be wise to try to unite them / abstract some of the boilerplate a bit.

### Update 2024/09/19

I have now added the following more finite field operations:

- `setZero`
- `cadd`
- `csub`
- `double
- `isZero`
- `isOdd`
- `shiftRight`
- `div2`

In addition, I split all the implementations for each finite field op into an internal and public part. That way we can reuse the operations in other ops.

Further, I added distinct `Array` types to deal with finite field points and elliptic curve points. With those, I then added a 'test case' of sorts, where I ported the EC short Weierstrass Jacobian `sumImpl` logic line by line. With the help of some templates the implementation is essentially the same as for the CPU now and the code produces the same result.

Next I'll clean up that test case and add the EC point addition to the `pub_curves.nim` module.

### Update 2024/09/23

I've now added the EC addition to a new `pub_curves.nim` module. In addition, I added a helper type `NvidiaAssembler` to simplify the Nvidia setup & compilation process. It can now be done with 2 lines:

```nim
  let nv = initNvAsm(EC_ShortW_Jac[field, G1], wordSize)  # either takes an EC_ShortW_Jac type or a Fp / Fr type
  let kernel = nv.compile(genEcSum) # pass the proc defining an `llvmPublicFnDef` 
```
(alternatively one can use the `Assembler_LLVM` part of the `NvidiaAssembler` object to build instructions / call a function and then just pass the name of the kernel to `compile`)

Using this, the boilerplate in the tests is reduced massively.

Note: The `t_ec_sum_port.nim` file is the one I used to write the port. I think it can be useful to give ideas on how to port some our existing CPU logic to the LLVM target.

Next: I'll add other EC operations.

### Update 2024/09/26

Over the last few days I have:
- split the `EcPoint` type into `EcPointAff`, `EcPointJac` and `EcPointProj`, which each have their own set of internal / public procedures
- added more misc. field and EC operations
- added templates `fieldOps`, `ellipticOps`, `ellipticAffOps` that inject multiple templates for operations to allow writing code equivalent to the CPU code that can 
- ported `mixedSum` for Jacobian + Affine EC points. Thanks to the templates, all that was needed to port the code was replace variable declarations and `=` assignments by `var X = asy.newEcPointJac(ed)` and `store(X, Y)`. Aside from a minor issue due to an `isNeutral` name collision (internal procs are still using `ValueRef` as arguments and hence overload resolution is an issue; since renamed the affine version), it worked immediately.


constantine/math_compiler/pub_fields.nim
```diff
@@ -6,12 +6,198 @@
...
+          asy.void_t, toTypes([r, a, b]),
+          {kHot}):
+    tagParameter(1, "sret")
+    let M = asy.getModulusPtr(fd)
+
+    let (ri, ai, bi) = llvmParams
+    asy.modadd(fd, ri, ai, bi, M)
+    asy.br.retVoid()
+
+  asy.callFn(name, [r, a, b])
```
mratsim, line 199: Ah, I see, there are 3 types of functions:
- underlying primitives
- internal
- public

and for fields, `internal` is just forwarding to modular arithmetic primitives

In the past I created an enum with the types of functions: https://github.com/mratsim/constantine/blob/cfe077b682b625b00d2e95693e8753570981d278/constantine/math_compiler/ir.nim#L202-L210

and I was planning to associate (opcode, curve) -> primitive, to call the proper underlying implementation.

I have yet to find the best solution but I think remove the _internal noise would be nice to improve debugging, perf optimization and dev experience.
Having all those `_internal` is annoying and unlike `_vartime` there is no security risk to justify it.

I think they should all be moved to the impl_*.nim file and renamed without `_internal`.

Public functions wont collide in LLVM due to the `ctt_` prefix and the Nim ones won't collide because the internal functions should not be reexported. Which is the case if they are implemented in the `pub_*.nim` like now.

constantine/math_compiler/pub_fields.nim
```diff
@@ -24,41 +210,879 @@ proc genFpAdd*(asy: Assembler_LLVM, fd: FieldDescriptor): string =
...
+    let rA = asy.asArray(ri, fd.fieldTy)
+    let aA = asy.asArray(ai, fd.fieldTy)
+    for i in 0 ..< fd.numWords:
+      rA[i] = aA[i]
+
+    # `r` now stores `a`
+    for i in countdown(count, 1):
+      # `mtymul` does not touch `r` until the end to store the result. It uses a temporary
+      # buffer internally. So we can just pass `r` 3 times!
+      asy.mtymul(fd, ri, ri, ri, M)
```
mratsim, line 580: All mtymul but the last can be called with `finalReduce = false`.
And internally mtymul must check whether there are at least 2 spare bits in the representation to allow this optimization.

This stems from the fact that Montgomery multiplication can actually accepts an input in the range [0, 2p) and outputs in [0, 2p) iff 4p < num words * word size (i.e. there are 2 spare bits for carries in the physical representation of the field).
Paper: https://eprint.iacr.org/2017/1057.pdf paragraph 2.4.1

There is a whole dance in the CPU impl about this:

https://github.com/mratsim/constantine/blob/cfe077b682b625b00d2e95693e8753570981d278/constantine/math/arithmetic/limbs_montgomery.nim#L501

constantine/math_compiler/pub_fields.nim
```diff
@@ -24,41 +210,879 @@ proc genFpAdd*(asy: Assembler_LLVM, fd: FieldDescriptor): string =
...
+    # from potentially larger values.
+    # So we create a boolean of whether it is 0, extend it to 32 bit and
+    # compute `0 - X` with `X ∈ {0, 1}`.
+
+    # create an int1 bool indicating if isZero is not zero: `notZero = isZero != 0`
+    let cZeroW = constInt(fd.wordTy, 0)
+    let notZero = asy.br.icmp(kNE, isZero, cZeroW)
+
+    # extend `int1_t` to `int32_t`
+    let notZeroExt = asy.br.zext(notZero, asy.ctx.int32_t())
```
mratsim, line 735: if we have to do this, we can just use the LLVM builtin select instead. Then the compiled should turn it in either conditional mov or predicated mov (or branches but not on a GPU).

constantine/math_compiler/pub_fields.nim
```diff
@@ -24,41 +210,879 @@ proc genFpAdd*(asy: Assembler_LLVM, fd: FieldDescriptor): string =
...
+  ## Generate a public field n-square proc
+  ## with signature
+  ##   `void name(FieldType r, FieldType a, count int)`
+  ## with `r` the resulting field element, `a` the field element to square
+  ## `count` times (i.e. for counts `[1, 2, 3, ...]` it yields `[a², a⁴, a⁸, ...]`).
+  ##
+  ## Returns the corresponding name to call it
+  let name = fd.name & "_nsqr"
+  let ptrBool = pointer_t(asy.ctx.int1_t())
+  ## XXX: int32? int64?
```
mratsim, line 858: `int32`, no one does 4 billions iterations of field squaring. RSA goes to 4096.
And it's way way faster on GPUs to use 32-bit numbers.

mratsim (commented):
Thank you. This in general looks good to me, especially the template that allows the GPU impl to not be overly verbose with the assembler context(s).

The `_internal` tag is a bit annoying though as it will be everywhere when debugging, optimizing or calling those functions while `pub` functions are only when calling the library and even then they are only generated and called in tests through their `ctt_` prefix. So I would move the internal functions in `impl`, maybe a new `impl_fields.nim` that just do the dispatch/wrapping and make sure that when reexporting `pub_fields.nim` users can't access internal functions.

There can be a macro that avoid the repetition of making public an impl function.

constantine/math_compiler/impl_fields_globals.nim
```diff
@@ -100,10 +100,175 @@ func checkValidModulus(M: BigNum) =
...
+  ## of non-secret data.
+  var borrow, diff: T
+  for i in 0 ..< a.limbs.len:
+    let ai = T(a.limbs[i])
+    let bi = T(b.limbs[i])
+    subB(borrow, diff, ai, bi, borrow)
+    if ctl:
+      a.limbs[i] = diff
+
+  result = bool borrow
```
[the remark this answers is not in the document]
Vindaar: The reason I did not use those is that they use statically sized arrays. It would require us to assign a static argument to many of the related procs that end up calling any of the above by needing any of the globals.

constantine/math_compiler/codegen_nvidia.nim
```diff
@@ -215,3 +218,321 @@ proc exec*[T](jitFn: CUfunction, r: var T, a, b: T) =
...
+
+proc getIdent(n: NimNode): NimNode =
+  ## Generate a `GPU` suffixed ident
+  doAssert n.kind in {nnkIdent, nnkSym}, "No, was: " & $n.treerepr
+  result = ident(n.strVal & "GPU")
+
+proc determineDevicePtrs(r, i: NimNode, iTypes: seq[NimNode]): seq[(NimNode, NimNode)] =
+  ## Returns the device pointer ident and its associated original symbol.
+  for el in r:
+    #if not el.requiresCopy:
```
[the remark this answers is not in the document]
Vindaar: I had some case where I triggered the error when I didn't expect to. Wanted to investigate at a later time and forgot and now I can't reproduce it. Will have another look and otherwise put it back in.

constantine/math_compiler/pub_fields.nim
Vindaar: Yup, indeed. The `finalReduce` wasn't exposed in the `mtymul` wrapper.

For the runtime variant for now I haven't added it though. It would require adding a version of `mtymul_nvidia` which can skip the reduce based on a runtime value. Can be done of course.

constantine/math_compiler/impl_curves_ops_jacobian.nim
```diff
@@ -0,0 +1,663 @@
...
+        Z2Z2 = asy.newField(cd.fd)
+        U2   = asy.newField(cd.fd)
+        S2   = asy.newField(cd.fd)
+
+      ## XXX: Similarly to `prod` below, `skipFinalSub = true` here, also breaks the
+      ## code at a different point.
+      Z2Z2.square(Q.z, skipFinalSub = false)
+      ## XXX: If we set this `skipFinalSub` to true, the code will fail to produce the
+      ## correct result in some cases. Not sure yet why. See `tests/gpu/t_ec_sum.nim`.
+      S1.prod(Q.z, Z2Z2, skipFinalSub = false)
```
Vindaar, line 284: Either of these cause problems in the code for multiple iterations in the mentioned test case. The two below (Z1Z1 and S2) work fine though. _Maybe_ we want to avoid it everywhere for the time being out of an abundance of caution though.

constantine/math_compiler/impl_curves_ops_jacobian.nim
```diff
@@ -0,0 +1,663 @@
...
+        U2 = asy.newField(cd.fd)
+        S2   = asy.newField(cd.fd)
+
+      U1 = P.x
+      S1 = P.y
+
+      ## XXX: either of these `skipFinalSub = true` breaks the code at some point.
+      ## See also `tests/gpu/t_ec_sum.nim`.
+      Z1Z1.square(P.z, skipFinalSub = false) #not (cd.coef_a == -3))
+      S2.prod(P.z, Z1Z1, skipFinalSub = false)
```
Vindaar, line 540: Same as above in the non mixed `sum` implementation.

constantine/math_compiler/codegen_nvidia.nim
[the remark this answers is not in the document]
Vindaar: Remembered the use case where it caused issues. I've now added test cases for `execCuda` and changed the logic to a `allowedCopy` for this. I.e. if one passes a `var` symbol as a result parameter, it is considered allowed (despite not being a `ptr T` type potentially). The macro logic now checks if it needs to apply an `addr` to the argument or not.

tests/gpu/t_mul.nim
```diff
@@ -0,0 +1,40 @@
...
+    kernel.execCuda(res = [rGPU], inputs = [a, b])
+
+    echo "CPU = ", rCPU.toHex()
+    echo "GPU = ", rGPU.toHex()
+    doAssert bool(rCPU == rGPU)
+
+let a = Fp[BN254_Snarks].fromUInt(1'u64)
+let b = Fp[BN254_Snarks].fromHex("0x2beb0d0d6115007676f30bcc462fe814bf81198848f139621a3e9fa454fe8e6a")
+
+testName(Fp[BN254_Snarks], 64, a, b)
```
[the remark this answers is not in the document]
mratsim: No issue on 32-bit, so same bug

tests/gpu/t_ec_sum_port.nim
```diff
@@ -0,0 +1,464 @@
...
+  constantine/math/io/[io_bigints, io_fields, io_ec],
+  constantine/math/arithmetic,
+  constantine/math/elliptic/ec_shortweierstrass_jacobian,
+  constantine/platforms/abstractions,
+  constantine/platforms/llvm/llvm,
+  constantine/math_compiler/[ir, pub_fields, pub_curves_jacobian, codegen_nvidia, impl_fields_globals],
+  # Test utilities
+  helpers/prng_unsafe
+
+proc genSumImpl*(asy: Assembler_LLVM, ed: CurveDescriptor): string =
```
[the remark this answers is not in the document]
mratsim: For now I've deleted it.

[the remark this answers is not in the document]
Vindaar: Yes, of sorts. It's the file of how I ported over the CPU version. I wanted to commit it so that it wouldn't be lost (in the PR branch at least!). Can be useful both for other people trying to get an idea on how to not get overwhelmed porting a more complex algorithm or simply myself in the future.

PR #470, opened 2024-09-25
fix: template typechecking is more stringent

mratsim:
Failing with latest `version-2-0` branch in Nim


mratsim:
The remaining bug is tracked in #471

PR #477, opened 2024-10-06
C bindings for Banderwagon

Richa-iitr:
Added C bindings for verkle-ipa.


include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+    const Fr_BanderWagon_OpenArray* opening_challenges_in_domain,
+    Fr_BanderWagon_OpenArray* evals_at_challenges,
+    IpaMultiProof_EcAff_Fr* proof
+    ) __attribute__((warn_unused_result));
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // __CTT_H_ETHEREUM_EVM_PRECOMPILES__
```
mratsim, line 228:
```suggestion
#endif // __CTT_H_ETHEREUM_VERKLE_IPA__
```

constantine/ethereum_verkle_ipa.nim
```diff
@@ -215,7 +217,7 @@ func deserialize*(dst: var EthVerkleIpaMultiProof,
 # TODO: refactor, this shouldn't use curves_primitives but internal functions
 import ./lowlevel_fields
 
-func mapToBaseField*(dst: var Fp[Banderwagon],p: EC_TwEdw[Fp[Banderwagon]]) =
+func mapToBaseField*(dst: var Fp[Banderwagon],p: EC_TwEdw[Fp[Banderwagon]]) {.libPrefix: prefix_ipa, discardable.} =
```
mratsim, line 220: Those should be renamed `map_to_base_field` so the C API uses proper snake-case. For existing Nim code it will have no regression due to being flexible on the casing.

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+    return statuses[status];
+  }
+  return "cttEthVerkleIpa_InvalidStatusCode";
+}
+
+// Opaque types for Nim-defined types
+typedef struct Fr Fr;
+typedef struct Banderwagon Banderwagon;
+typedef struct EC_TwEdw EC_TwEdw;
+typedef struct Fp Fp;
```
mratsim, line 54: All the fields and curves related struct should be defined in `constantine/include/constantine/curves/banderwagon.h`

This would be similar to pallas and vesta as they don't use pairings: https://github.com/mratsim/constantine/blob/2c3de8738d031394dfb78e3fc839cf5d12e42b92/bindings/lib_curves.nim#L68-L100

However Banderwagon is an Edwards curve so the template there might need an adaptation.
I don't remember what I add in mind as differences between short-weierstrass and twisted edwards, maybe it was just the affine<->jacobian<->projective coordinate conversion and that could be done as a separate template, see https://github.com/mratsim/constantine/blob/2c3de8738d031394dfb78e3fc839cf5d12e42b92/bindings/c_curve_decls.nim#L314-L432

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef union {
+    uint8_t u8;
+    uint16_t u16;
+    uint32_t u32;
+    uint64_t u64;
+    unsigned int u;
+} SomeUnsignedInt;
```
mratsim, line 62: You cannot emulate generics with union types.

Whatever the platform, the storage reserved will be 64-bit / 8 bytes per entry here.
But if Nim uses 32-bit inputs in an array a, the C code will read both entries with `a[0]` and will read past the buffer with `a[1]`.

Hence generic procedures cannot be wrapped in C directly.

constantine/commitments/eth_verkle_ipa.nim
```diff
@@ -139,7 +143,7 @@ func innerProduct[F](r: var F, a, b: distinct(View[F] or MutableView[F])) =
 func ipa_commit*[N: static int, EC, F](
       crs: PolynomialEval[N, EC],
       r: var EC,
-      poly: PolynomialEval[N, F]) =
+      poly: PolynomialEval[N, F]) {.libPrefix: prefix_ipa.} =
```
mratsim, line 146: This is not the proper level of exports as this is a generic procedure.

Export should happen at fully specified protocol level (i.e. curve, field, hash functions, `N: static int` input sizes are known).

In this case, this means having proper instantiated `ipa_commit` here: 
https://github.com/mratsim/constantine/blob/2c3de8738d031394dfb78e3fc839cf5d12e42b92/constantine/ethereum_verkle_ipa.nim#L285-L292

and tagging those with libPrefix.

constantine/commitments/eth_verkle_ipa.nim
```diff
@@ -330,7 +334,7 @@ func ipa_verify*[N, logN: static int, EcAff, F](
       commitment: EcAff,
       opening_challenge: F,
       eval_at_challenge: F,
-      proof: IpaProof[logN, EcAff, F]): bool =
+      proof: IpaProof[logN, EcAff, F]): bool {.libPrefix: prefix_ipa.} =
```
mratsim, line 337: ditto

constantine/commitments/eth_verkle_ipa.nim
```diff
@@ -661,7 +665,7 @@ func ipa_multi_prove*[N, logN: static int, EcAff, F](
       proof: var IpaMultiProof[logN, EcAff, F],
       polys: openArray[PolynomialEval[N, F]],
       commitments: openArray[EcAff],
-      opening_challenges_in_domain: openArray[SomeUnsignedInt]) =
+      opening_challenges_in_domain: openArray[SomeUnsignedInt]) {.libPrefix: prefix_ipa.} =
```
mratsim, line 668: ditto

constantine/commitments/eth_verkle_ipa.nim
```diff
@@ -838,7 +842,7 @@ func ipa_multi_verify*[N, logN: static int, EcAff, F](
       commitments: openArray[EcAff],
       opening_challenges_in_domain: openArray[SomeUnsignedInt],
       evals_at_challenges: openArray[F],
-      proof: IpaMultiProof[logN, EcAff, F]): bool =
+      proof: IpaMultiProof[logN, EcAff, F]): bool {.libPrefix: prefix_ipa.} =
```
mratsim, line 845: ditto

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+typedef struct {
+    SomeUnsignedInt* values;  
+    size_t length;
+} SomeUnsignedInt_Array;
+
+// Define types for openArray
+typedef struct {
+    Fr* data;    
+    size_t len;   
+} Fr_BanderWagon_OpenArray;
```
mratsim, line 73: This is not needed, you can pass directly ptr + len.

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct {
+    EC_TwEdw* data;    
+    size_t len;         
+} EC_TwEdw_Fp_Banderwagon_OpenArray;
```
mratsim, line 78: not needed

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct {
+    byte value[32];  // 32-byte array for field element
+} Fp_Banderwagon;
```
mratsim, line 82: should be autogenerated in `constantine/include/constantine/curves/banderwagon.h`

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct {
+    byte value[32];  // 32-byte array for field element
+} Fr_Banderwagon;
```
mratsim, line 86: ditto

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct {
+    byte x[32];  // 32-byte x-coordinate
+    byte y[32];  // 32-byte y-coordinate
+} EC_TwEdw_Fp_Banderwagon;
```
mratsim, line 91: ditto

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct {
+    byte x[32];  // 32-byte x-coordinate
+    byte y[32];  // 32-byte y-coordinate
+} EC;
```
mratsim, line 96: ditto

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct {
+    EC* points;  
+    size_t length;  
+} PolynomialEval_EC;
```
mratsim, line 101: the protocol has fixed size polynomials. Nim `static int` are similar to generics and cannot be represented with C runtime parameters. Instead a no generic/no static procedure should be implemented at the croot of the project in `constantine/ethereum_verkle_ipa.nim`

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+typedef struct {
+    EC* ec_points;  
+    Fr* field_elements;      
+    size_t logN;                         
+} IpaMultiProof_EcAff_Fr;
+
+typedef struct {
+    EC* ec_points;  
+    size_t length;                       
+} EcAffArray;
```
mratsim, line 133: ditto for all structs up to here

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,228 @@
...
+
+typedef struct EthVerkleIpaProof EthVerkleIpaProof;
+typedef struct EthVerkleIpaMultiProof EthVerkleIpaMultiProof;
+typedef struct IpaProof IpaProof;
+typedef struct IpaMultiProof IpaMultiProof;
+typedef struct EthVerkleTranscript EthVerkleTranscript;
+
+typedef byte EthVerkleIpaProofBytes[544];       // Array of 544 bytes
+typedef byte EthVerkleIpaMultiProofBytes[576];
```
mratsim, line 142: It's more useful to provide:
- either `ctt_eth_verkle_ipa_proof_sizeof()` and `ctt_eth_verkle_ipa_multiproof_sizeof()` and let the caller do their allocation / destruction
- or `ctt_eth_verkle_ipa_proof_create()` / `ctt_eth_verkle_ipa_proof_destroy()`
  and `ctt_eth_verkle_ipa_multiproof_create()` / `ctt_eth_verkle_ipa_multiproof_destroy()`

mratsim (commented):
Thanks for the contributions.

I think this should be split into 3 steps:
1. C bindings for banderwagon fr, fp, ec
2. Creating a proper wrapper for the Ethereumn specific instantiation of IPA. Currently the generic code is forwarded (i.e. `constantine/commitments/eth_verkle_ipa.nim`) but protocol-specific constants/types should be wrapped and exposed in a proper procedure in the root `constantine/eth_verkle_ipa.nim`  and only there can C bindings be generated from.
3. Actually do the C bindings.

include/constantine/protocols/ethereum_verkle_ipa.h
Richa-iitr: @mratsim  i resolved this and added separate header for banderwagon. Please review and suggest changes. Also how can i verify things are working as expected?

constantine/commitments/eth_verkle_ipa.nim
Richa-iitr: is there any example of a wrapper i can follow to understand?

mratsim: Yes, anything with generics cannot be exported and need a concretely typed wrapper.

include/constantine/protocols/ethereum_verkle_ipa.h
mratsim: Good question, I don't have a testsuite for the C API of elliptic curves. It's not ideal but for now we can skip it merge as-is and focus on the C API for the full protocol.

mratsim (approved):
The PR looks good for a merge if we remove the `protocols/ethereum_verkle_ipa.h`. Its rewrite can be done separately.

Richa-iitr:
> The PR looks good for a merge if we remove the `protocols/ethereum_verkle_ipa.h`. Its rewrite can be done separately.

removed the file, please review

Richa-iitr:
Hi @mratsim 
I am done with the bindings, left with a few c test scenarios which would complete in the next few days. There's one issue i am facing, the tests are passing in my linux system with nim version 2.2.0 as is in the ci,
image
But the tests are failing in the CI, 
image
Is there something i am missing here?

constantine/ethereum_verkle_ipa.nim
```diff
@@ -137,10 +157,19 @@ template check(Section: untyped, evalExpr: CttCodecEccStatus): untyped {.dirty.}
...
+    aff: IpaProof[8, EC_TwEdw_Aff[Fp[Banderwagon]], Fr[Banderwagon]]
+    prj: IpaProof[8, EC_TwEdw_Prj[Fp[Banderwagon]], Fr[Banderwagon]]
+  EthVerkleIpaMultiProof* {.exportc: prefix_ffi & "multi_proof".} = object
+    aff: IpaMultiProof[8, EC_TwEdw_Aff[Fp[Banderwagon]], Fr[Banderwagon]]
+    prj: IpaMultiProof[8, EC_TwEdw_Prj[Fp[Banderwagon]], Fr[Banderwagon]]
+  EthVerkleIpaPolyEvalCrs* {.exportc: prefix_ffi & "poly_eval".} = object
+    aff: PolynomialEval[EthVerkleDomain, EC_TwEdw_Aff[Fp[Banderwagon]]]
+    prj: PolynomialEval[EthVerkleDomain, EC_TwEdw_Prj[Fp[Banderwagon]]]
+  EthVerkleIpaPolyEvalPoly* {.exportc: prefix_ffi & "poly_eval_poly".} = object
+    raw: PolynomialEval[256, Fr[Banderwagon]]
```
mratsim, line 172: The file should be reverted and not modified.

This approach doubles object size while we only need aff or prj coordinates. The serialized code should just use `EthVerkleIpaProofBytes` and unsure yet on the C API of the internal proof type, maybe will hide it behind an opaque pointer.

constantine/ethereum_verkle_ipa.nim
```diff
@@ -167,8 +196,20 @@ func serialize*(dst: var EthVerkleIpaProofBytes,
...
+                src: EthVerkleIpaProof
+                ): cttEthVerkleIpaStatus {.libPrefix: prefix_ffi.} =
+  
+  return serialize(dst, src.aff)
+
+func serialize_prj*(dst: var EthVerkleIpaProofBytes,
+                src: EthVerkleIpaProof
+                ): cttEthVerkleIpaStatus {.libPrefix: prefix_ffi.} =
+  
+  return serialize(dst, src.prj)
```
mratsim, line 209: A user will not know what serialization to use between prj and aff, this will lead to errors and so is not the correct API.

examples-c/ethereum_verkle_ipa.c
```diff
@@ -0,0 +1,60 @@
...
+  //       printf("%02X ", proof_bytes.raw[i]);
+  //       if ((i + 1) % 16 == 0) {
+  //           printf("\n"); // Add a newline for readability after every 16 bytes
+  //       }
+  //   }
+  //   printf("\n");
+  ctt_eth_verkle_ipa_proof_aff proof;
+
+  ctt_eth_verkle_ipa_status stat = ctt_eth_verkle_ipa_deserialize_aff(&proof, &proof_bytes);
```
mratsim, line 55: File to delete so the PR is focused on Banderwagon

include/constantine/protocols/ethereum_verkle_ipa.h
```diff
@@ -0,0 +1,134 @@
+/** Constantine
+ *  Copyright (c) 2018-2019    Status Research & Development GmbH
+ *  Copyright (c) 2020-Present Mamy André-Ratsimbazafy
+ *  Licensed and distributed under either of
+ *    * MIT license (license terms in the root directory or at http://opensource.org/licenses/MIT).
+ *    * Apache v2 license (license terms in the root directory or at http://www.apache.org/licenses/LICENSE-2.0).
+ *  at your option. This file may not be copied, modified, or distributed except according to those terms.
+ */
+#ifndef __CTT_H_ETHEREUM_VERKLE_IPA__
+#define __CTT_H_ETHEREUM_VERKLE_IPA__
```
mratsim, line 10: File to delete so the PR is focused on Banderwagon / twisted edwards

tests/t_ethereum_verkle_ipa_primitives.nim
```diff
@@ -760,12 +760,12 @@ suite "Multiproof Tests":
 
       ## Deserialize it into the Multiproof type
       var multiproof {.noInit.}: IpaMultiProof[8, EC_TwEdw_Aff[Fp[Banderwagon]], Fr[Banderwagon]]
-      let s1 = multiproof.deserialize(validMultiproof_bytes)
+      let s1 = multiproof.batch_deserialize(validMultiproof_bytes)
```
mratsim, line 763: revert

mratsim (changes requested):
As mentioned in my previous review (https://github.com/mratsim/constantine/pull/477#pullrequestreview-2351247910), the PR needs to be separated in elliptic curve API which are well documented and a Verkle API.

For now, let's delete all the Verkle API changes as they cannot be used as-is and I am unsure myself on how to design the API. Furthermore the future of the Verkle Tree upgrade is in flux at the moment.

Richa-iitr:
@mratsim  i have reverted the verkle api changes, the pr is ready for review.

constantine/commitments/eth_verkle_ipa.nim
```diff
@@ -968,4 +964,4 @@ func ipa_multi_verify*[N, logN: static int, EcAff, F](
   return crs.ipa_verify(
     domain, transcript,
     comm_g2_aff, t,
-    g2t, proof.g2_proof)
+    g2t, proof.g2_proof)
```
mratsim, line 967:
```suggestion
    g2t, proof.g2_proof)

```

constantine/ethereum_verkle_ipa.nim
```diff
@@ -289,4 +289,4 @@ func batchMapToScalarField*(
 # - sha256 for transcripts
 
 export eth_verkle_ipa
-export hashes
+export hashes
```
mratsim, line 292:
```suggestion
export hashes

```

examples-c/t_libctt_banderwagon.c
```diff
@@ -0,0 +1,234 @@
...
+  printf(" SUCCESS Legendre symbol / is_square\n");
+
+  mpz_clear(r);
+  mpz_clear(p);
+  mpz_clear(b);
+  mpz_clear(a);
+
+  return 0;
+
+}
```
mratsim, line 234:
```suggestion
}

```

tests/t_ethereum_verkle_ipa_primitives.nim
```diff
@@ -776,4 +776,4 @@ suite "Multiproof Tests":
 # TODO - missing tests from:
 # - https://github.com/crate-crypto/verkle-trie-ref/blob/2332ab8/multiproof/multiproof_test.py
 # - https://github.com/crate-crypto/go-ipa/blob/b1e8a79/multiproof_test.go
-# - https://github.com/crate-crypto/rust-verkle/blob/442174e/ipa-multipoint/src/multiproof.rs
+# - https://github.com/crate-crypto/rust-verkle/blob/442174e/ipa-multipoint/src/multiproof.rs
```
mratsim, line 779:
```suggestion
# - https://github.com/crate-crypto/rust-verkle/blob/442174e/ipa-multipoint/src/multiproof.rs

```

mratsim (approved):
LGTM, merging after the files have an extra empty line

bindings/lib_curves.nim
```diff
@@ -140,3 +142,16 @@ collectBindings(cBindings_vesta_parallel):
...
+  banderwagon_fp = Fp[Banderwagon]
+  banderwagon_ec_aff = EC_TwEdw_Aff[Fp[Banderwagon]]
+  banderwagon_ec_prj = EC_TwEdw_Prj[Fp[Banderwagon]]
+
+collectBindings(cBindings_banderwagon):
+  genBindingsField(big253, banderwagon_fr)
+  genBindingsField(big255, banderwagon_fp)
+  genBindingsFieldSqrt(banderwagon_fp)
+  genBindings_EC_TwEdw_Affine(banderwagon_ec_aff, banderwagon_fp)
+  genBindings_EC_TwEdw_Projective(banderwagon_ec_prj, banderwagon_ec_aff, big253, banderwagon_fr)
```
mratsim, line 157:
```suggestion
  genBindings_EC_TwEdw_Projective(banderwagon_ec_prj, banderwagon_ec_aff, big253, banderwagon_fr)

```

PR #484, opened 2024-11-27
fix(MSM): properly handle edge condition in parallel MSM when bits is exactly divided by c

mratsim:
The parallel MSM skips the top window if the number of bits is exactly divided by MSM parameter c.

This is easy to reproduce with 3 points and by removing the endomorphism acceleration (which multiply the number of points by 2x or 4x) here:

https://github.com/mratsim/constantine/blob/7c3d76d6d6e4de78f3d5a4d37fe39ed4b247c491/constantine/math/elliptic/ec_multi_scalar_mul_parallel.nim#L539-L557

This might help for #479, cc @Vindaar


mratsim:
unrelated failure `go test` on Windows

image

mratsim:
image

PR #445, opened 2024-07-27
Crandall primes

mratsim:
This closes #11 for primes of form 2ᵐ-c (Crandall primes / pseudo-Mersenne primes), such as the one used for Curve25519 and secp256kq (Ethereum/ Bitcoin).

### Previous

image
image

### Current

image
image

### Analysis

- Fp[Edwards25519] mul 1.27x improvement
- Fp[Edwards25519] square 1.43x improvement
- Fp[Secp256k1] mul 1.94x improvement
- Fp[Secp256k1] square 1.28x improvement

### Previous

image

### Current

image

### Analysis

- EC add projective constant-time improved by 1.36x
- EC add jacobian constant-time improved by 1.34x
- EC add projective vartime improved by 1.24x
- EC add jacobian vartime improved by 1.37x
- EC dbl projective constant-time improved by 1.31x
- EC dbl jacobian constant-time improved by 1.06x

## Bench vs bitcoin/secp256k1

image
- field_sqr 12.4ns vs 8ns -> 1.55x
- field_mul 15.8ns vs 10ns -> 1.58x
- field_inv_ct 1410ns vs 1203ns -> 1.17x
- field_inv_vt 820ns vs 848ns -> 0.97x
- EC add jacobian var 247ns vs 97ns -> 2.55x
- EC dbl jacobian var 97.8ns vs 145 -> 0.67x
- EC mixed add ct 189ns vs 225ns -> 0.84x
- EC mixed add var 173ns vs 98ns -> 1.77x

image

- EC scalar-mul ct  28100ns vs  40196 ns -> 0.70x

## Analysis

The fact that field operations are 1.5x faster BUT the elliptic curve operations are sometimes slower is suspicious. We probably need to check the EC formulae

## TODO

- fix windows
- bound checks for lazy reduce and lazy reduced field exponentiation for 256-bit as eprint/iacr 2018/985
  indicates in Theorem 4 that their partial reduction may grow by 1 bit if 256-bit.
- optimize EC impl to avoid if/else check for ADX and limit input/output movement
- optimized mixed add

PR #485, opened 2024-11-27
Torus-acceleration for multiexponentiation on GT

mratsim:
This project was sponsored by the Ethereum Foundation under the name
_Implementing & Accelerating Torus-based cryptography for SSLE_, FY24-1672
and is a collaboration with Robert Granger (https://www.surrey.ac.uk/people/robert-granger) and Antonio Sanso (@asanso).

## Overview

Currently, Ethereum validators are known up to 6.4 minutes in advance (1 Epoch = 32 slots = 32*12 seconds).
In theory a malicious actor may do denial-of-service attacks on upcoming validators to prevent block production.

To prevent this a Single Secret Leader Election protocol (SSLE) may be used to keep the identity of a block producer under wrap.

The protocol identified for Ethereum is Whisk:
- https://ethresear.ch/t/whisk-a-practical-shuffle-based-ssle-protocol-for-ethereum/11763
  - https://github.com/ethereum/consensus-specs/pull/2800, https://github.com/ethereum/consensus-specs/pull/3205
  - https://github.com/asn-d6/curdleproofs/blob/1aa27a6/doc/curdleproofs.pdf  

However it was initially instantiated on elliptic curve groups (G1 or G2) and pairings could unravel the whole scheme (similar to I guess the MOV attack https://crypto.stanford.edu/pbc/notes/elliptic/movattack.html ).
To prevent this attack vector, the scheme can be instantiated on the pairing group GT.

This means computations on Fp12 instead of G1 (2x Fp), meaning we should expect computations to be 6x more expensive.

Full overview: https://crypto.stanford.edu/pbc/notes/elliptic/movattack.html

The goal of this PR is to reduce the overhead of GT multiexponentiation from the initial 5x https://ethresear.ch/t/the-return-of-torus-based-cryptography-whisk-and-curdleproof-in-the-target-group/16678/3, ideally to 3x to make Whisk viable latency-wise.

## Implementation

- This adds torus-based cryptography for GT that can be combined with 4-way endomorphism acceleration
  - It uses projective coordinates to delay "affine" conversion, similar to elliptic curve affine/projective coordinates 
- It adds a Toom-Cook+DFT multiplication/squaring algorithm for Fp6 (unused at the moment as unexpectedly slower, see `Future work` section

Note: the Torus acceleration as implemented seems to be only valid for curve with 1+𝑖 as sextic non-residue, with 𝑖 = √-1. This is the case for BLS12-381 and BN254-Nogami but not for BN254-Snarks (Zcash, Ethereum) or BLS12-377 (Aleo)

## Benchmarks

We compare to BLST MSM G1, BLST is used by all consensus clients today, benchmarks from https://github.com/status-im/nim-blscurve/pull/183.

The multiexp size is expected to be 128 or 256. We implement using 2 towering schemes. Note that Torus acceleration is only valid for Fp12 over Fp6 over Fp2

Machine is a Ryzen 9 9950XE, overclocked at 5.9GHz single-threaded

BLST MSM G1
image

Constantine MEXP GT
image

**For 128 points, the ratio G1/GT is only 3x.**
For 256 points, the ratio G1/GT is 3.28x.

## Future work

I have been investigating some performance bugs in the summer where Constantine starts with a 1.7x perf advantage in field arithmetic that get reduced to 1x at the elliptic curve level or even worse after Fp2 -> Fp6 -> Fp12 towering:
- https://github.com/mratsim/constantine/pull/445#issuecomment-2254191531
- https://github.com/mratsim/constantine/pull/462

Solving this might reduce the gap to 2.4x (20% improvement)

Another line of work is to use SIMD for multiexp to compute on 4x 64-bit integers (AVX2) or 8x 64-bit integers (AVX512) per instruction which should conservatively bring at least a 2x / 4x perf improvement respectively.


burdges:
Why not use Ristretto or similar for Whisk?  

We've a similar issue in the ring VRFs used by Sassafras on Polkadot, since my optimized design has VUF outputs on G1, but then you could move them onto a "sister curve", ala https://github.com/cfrg/draft-irtf-cfrg-bls-signature/issues/30.  Alistair later noticed a Plonk-ish ring VRF is plenty fast enough for our validator set size, which used VUF outputs on bandersnatch, so no requirement.  I suppose a Risttretto bullet-proof ring VRF maybe even faster for us too.  Afaik none of that helps you, not without replacing Whick by Sassafras.

asanso:
> Why not use Ristretto or similar for Whisk?
> 
> We've a similar issue in the ring VRFs used by Sassafras on Polkadot, since my optimized design has VUF outputs on G1, but then you could move them onto a "sister curve", ala [cfrg/draft-irtf-cfrg-bls-signature#30](https://github.com/cfrg/draft-irtf-cfrg-bls-signature/issues/30). Alistair later noticed a Plonk-ish ring VRF is plenty fast enough for our validator set size, which used VUF outputs on bandersnatch, so no requirement. I suppose a Risttretto bullet-proof ring VRF maybe even faster for us too. Afaik none of that helps you, not without replacing Whick by Sassafras.

because as highlighted in https://ethresear.ch/t/the-return-of-torus-based-cryptography-whisk-and-curdleproof-in-the-target-group/16678 we want to take the easy way and use the validator’s secret signing key k and its associated public key kG1 for bootstrapping.

PR #486, opened 2024-11-27
CI: drop old nim compiler versions

### LINKED ISSUE #471: Compiler crash in nim/nightlies
The nightly currently running in CI is 9d65507d202afcfdf9421fc1e2c9cba3325059ba

and crashes with `Error: unhandled exception: int128.nim(69, 11) `arg.sdata(3) == 0` out of range [AssertionDefect]`

The latest Nim commit in `devel` doesn't crash, at least on my machine.

mratsim:
We remove explicit support for Nim 1.6.x as its EOL and Nim 2.0.x as one version to the next works or does not work.

Furthermore neither 1.6.12 or 2.0.12 can compile #485

Closes #471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,10 @@ jobs:
       fail-fast: false
       max-parallel: 20
       matrix:
-        nim_version: [version-1-6, version-2-2] # version-2-0] is crashing due to https://github.com/mratsim/constantine/issues/471
+        # version-2-0 is crashing due to https://github.com/mratsim/constantine/issues/471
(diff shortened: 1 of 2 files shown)


tersec:
Minimal reproducer for this in Nim 2.0 series:
```nim
template y(_: static int): untyped = int
type S[F: static int] = y(F)
type E[C: static int] = S[C]
discard sizeof(E[0])
```

narimiran:
FWIW, a workaround is to change `y(F)` to `y F` or `F.y`.

@mratsim, would you be interested in making those changes to make `constantine` compatible with Nim 2.0.x again?

PR #489, opened 2024-12-01
Add KZG point precompile

Vindaar:
This adds the KZG point precompile:

https://eips.ethereum.org/EIPS/eip-4844#point-evaluation-precompile

There are several points I'm not sure about:
- the input is given in big endian order and the Python spec simply slices parts of it and works with those. While the SHA256 hash of the commitment is given explicitly (and thus should be done in big endian order afaict), it's unclear to me how to handle it for `verify_kzg_proof`. The latter works on byte arrays. While I'd normally assume those would be in little endian order, I don't know. So for now I just copy the data like in the spec in big endian order.
- due to adding it to the EVM precompiles file and not the KZG file, we lack some sensible enum values in the `CttEVMStatus` enum. Do we add e.g. `cttEVM_KzgVerificationFailure`?
- the spec mentions test cases https://eips.ethereum.org/EIPS/eip-4844#test-cases, but while there is a Python file for KZG point precompile tests https://github.com/ethereum/execution-spec-tests/blob/1983444bbe1a471886ef7c0e82253ffe2a4053e1/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py in the test vectors I don't see anything specific to the precompile. 

Otherwise, once this is sane, I'd go in and add the
- [ ] C API
- [ ] Go wrapper
- [ ] Rust wrapper


mratsim:
> the input is given in big endian order and the Python spec simply slices parts of it and works with those. While the SHA256 hash of the commitment is given explicitly (and thus should be done in big endian order afaict), it's unclear to me how to handle it for verify_kzg_proof. The latter works on byte arrays. While I'd normally assume those would be in little endian order, I don't know. So for now I just copy the data like in the spec in big endian order.

`verify_kzg_proof` is specified in Consensus spec https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md#verify_kzg_proof
```python
def verify_kzg_proof(commitment_bytes: Bytes48,
                     z_bytes: Bytes32,
                     y_bytes: Bytes32,
                     proof_bytes: Bytes48) -> bool:
    """
    Verify KZG proof that ``p(z) == y`` where ``p(z)`` is the polynomial represented by ``polynomial_kzg``.
    Receives inputs as bytes.
    Public method.
    """
    assert len(commitment_bytes) == BYTES_PER_COMMITMENT
    assert len(z_bytes) == BYTES_PER_FIELD_ELEMENT
    assert len(y_bytes) == BYTES_PER_FIELD_ELEMENT
    assert len(proof_bytes) == BYTES_PER_PROOF

    return verify_kzg_proof_impl(bytes_to_kzg_commitment(commitment_bytes),
                                 bytes_to_bls_field(z_bytes),
                                 bytes_to_bls_field(y_bytes),
                                 bytes_to_kzg_proof(proof_bytes))
```

and this also defines the `Bytes32` and `Bytes48` endianness as big in https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md#constants

with the constant `KZG_ENDIANNESS`

In general, cryptographic protocols use big endian for interop it's much nicer to debug bigint that way. And internal representation is little-endian because it's consistent with popular platforms and also much more natural to implement all operations in little-endian (especially for carry/borrow propagation).

> due to adding it to the EVM precompiles file and not the KZG file, we lack some sensible enum values in the CttEVMStatus enum. Do we add e.g. cttEVM_KzgVerificationFailure?

Yes, as the last value so we don't break existing users.

> but while there is a Python file for KZG point precompile tests

This is a test generator, the test vectors can be downloaded with something similar to the script here: https://github.com/status-im/nim-eth2-scenarios (this is for consensus while those are execution layer test vectors).

But the simplest should be to pick the one from Geth: https://github.com/ethereum/go-ethereum/blob/ce8cec007c42e54f79e53c3769a7279487968c07/core/vm/testdata/precompiles/pointEvaluation.json

constantine/ethereum_evm_precompiles.nim
```diff
@@ -1215,3 +1218,60 @@ func eth_evm_bls12381_map_fp2_to_g2*(r: var openArray[byte], inputs: openarray[b
...
+  if len(r) != 64:
+    return cttEVM_InvalidOutputSize
+
+  var
+    versioned_hash: array[32, byte]
+    z: array[32, byte]
+    y: array[32, byte]
+    commitment_bytes: array[48, byte]
+    proof: array[48, byte]
+  ## XXX: copy like Python code in big endian order or unmarshal? `verify_kzg_proof` works on byte arrays
```
mratsim, line 1252: https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md#constants

`verify_kzg_proof` is bigendian

constantine/ethereum_evm_precompiles.nim
```diff
@@ -1215,3 +1218,60 @@ func eth_evm_bls12381_map_fp2_to_g2*(r: var openArray[byte], inputs: openarray[b
...
+  z.rawCopy(0, input, 32, 32)
+  y.rawCopy(0, input, 64, 32)
+  commitment_bytes.rawCopy(0, input, 96, 48) # commitment_bytes should surely be hashed in big endian order based on spec
+  proof.rawCopy(0, input, 144, 48)
+
+  # Verify commitment matches versioned_hash
+  var rhash: array[32, byte]
+  rhash.kzg_to_versioned_hash(commitment_bytes)
+  if rhash != versioned_hash:
+    ## XXX: BAD! If it was `cttEthKzgStatus` return type I'd return the VerificationFailure
```
mratsim, line 1263: let's add a new enum value

constantine/ethereum_evm_precompiles.nim
```diff
@@ -1215,3 +1218,60 @@ func eth_evm_bls12381_map_fp2_to_g2*(r: var openArray[byte], inputs: openarray[b
...
+    return cttEVM_InvalidInputSize #cttEthKzg_VerificationFailure
+
+  # Verify KZG proof with z and y in big endian format
+  var ctx = allocHeapAligned(EthereumKZGContext, alignment = 64)
```
mratsim, line 1267: You need to use `trusted_setup_load`

constantine/ethereum_evm_precompiles.nim
```diff
@@ -1215,3 +1218,60 @@ func eth_evm_bls12381_map_fp2_to_g2*(r: var openArray[byte], inputs: openarray[b
...
+  if ctx.verify_kzg_proof(commitment_bytes, z, y, proof) != cttEthKzg_Success:
+    ## XXX: BAD! If it was `cttEthKzgStatus` return type I'd return the VerificationFailure
+    return cttEVM_InvalidInputSize #cttEthKzg_VerificationFailure
+
+  # Reference:
+  # `return Bytes(U256(FIELD_ELEMENTS_PER_BLOB).to_be_bytes32() + U256(BLS_MODULUS).to_be_bytes32())`
+  r.toOpenArray( 0, 32-1).marshal([FIELD_ELEMENTS_PER_BLOB], WordBitWidth, bigEndian)
+  r.toOpenArray(32, 64-1).marshal(Fp[BLS12_381].getModulus(), bigEndian)
+
+  result = cttEVM_Success
```
mratsim, line 1277: There is no corresponding dealloc (well should be `trusted_setup_delete`

mratsim (commented):
Beyond the inline comments, the precompile should take a trusted setup as input as well so it can be cached by the Ethereum Execution Client.

constantine/ethereum_evm_precompiles.nim
Vindaar: Duh, right. :facepalm:

constantine/ethereum_evm_precompiles.nim
Vindaar: Given that I now have a parameter for a `EthereumKZGContext` this is only needed for the calling code, no? Or rather, we don't intend to add an overload, which only takes a filename to a trusted setup file, I imagine?

Vindaar:
Test case is added and the precompile now takes an `EthereumKZGContext` as first argument.

Do we wish to change the name of the proc in some way? I added the `eth_evm` prefix, but kept the `point_evaluation_precompile` from the spec as a suffix. We could shorten that of course.

Maybe we do the changes to the C API and Go, Rust wrappers in a separate PR after this?

constantine/ethereum_evm_precompiles.nim
mratsim: yes indeed

mratsim (dismissed):
looks good.

Yes for the name we can remove the `_precompile` suffix.

And agree we can do Rust and Go in a separate PR

tests/t_ethereum_evm_precompiles.nim
```diff
@@ -34,16 +34,37 @@ const
...
+    "trusted_setup_ethereum_kzg4844_reference.dat"
+
 proc loadVectors(TestType: typedesc, filename: string): TestType =
   let content = readFile(TestVectorsDir/filename)
   result = content.fromJson(TestType)
 
-template runPrecompileTests(filename: string, funcname: untyped, outsize: int) =
+proc trusted_setup*(): ptr EthereumKZGContext =
+  ## This is a convenience function for the Ethereum mainnet testing trusted setups.
+  ## It is insecure and will be replaced once the KZG ceremony is done.
```
mratsim, line 50: actually the trusted setup being loaded now is the official one from the KZG ceremony
```suggestion
```

tests/t_ethereum_evm_precompiles.nim
```diff
@@ -34,16 +34,37 @@ const
...
+    "commitments_setups" /
+    "trusted_setup_ethereum_kzg4844_reference.dat"
+
 proc loadVectors(TestType: typedesc, filename: string): TestType =
   let content = readFile(TestVectorsDir/filename)
   result = content.fromJson(TestType)
 
-template runPrecompileTests(filename: string, funcname: untyped, outsize: int) =
+proc trusted_setup*(): ptr EthereumKZGContext =
+  ## This is a convenience function for the Ethereum mainnet testing trusted setups.
```
mratsim, line 49:
```suggestion
```

mratsim (dismissed):

PR #490, opened 2024-12-12
Add ECDSA over secp256k1 signatures and verification

Vindaar:
This adds support for the Elliptic Curve Digital Signature Algorithm (ECDSA) defined over secp256k1. 

The current API supports:
- [X] sample random private keys 
- [X] derive their public keys
- [X] sign a message
- [X] verify a message
- [X] convert a signature into ASN.1 DER encoding following SEC1
- [X] write PEM files for private and public keys (NOTE: currently 'unclean' in the sense that they partially depend on the stdlib. Main dependency is `base64` encoding)
- [X] perform lower-s normalization for signatures (maybe we want an option to disable?)
- [X] generate nonces via
  - [X] uniform random sampling
  - [X] deterministic nonces based on [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979#section-3.2)
  
Currently still missing:
- [ ] batch verification
- [ ] ~~public key recovery~~ ⇒ will be part of PR #504 
- [x] more test cases (currently: script to generate signatures using OpenSSL and verify those with Constantine). Maybe sign messages with CTT & verify using OpenSSL

In addition, there are a few minor things I'm unsure about. Will highlight them below.


constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+  dst.fromBig(scalar)
+
+proc randomFieldElement[FF](): FF =
+  ## random element in ~Fp[T]/Fr[T]~
+  let m = FF.getModulus()
+  var b: matchingBigInt(FF.Name)
+
+  while b.isZero().bool or (b > m).bool:
+    ## XXX: raise / what else to do if `sysrand` call fails?
+    doAssert b.limbs.sysrand()
```
Vindaar, line 89: In case this were to fail, do we want a different failure mode?

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+    # Generate random nonce
+    var k = generateNonce(nonceSampler, message_hash, privateKey)
+
+    # Calculate r (x-coordinate of kG)
+    # `r = k·G (mod n)`
+    let r_point = k * G
+    # get x coordinate of the point `r` *in affine coordinates*
+    let rx = r_point.getAffine().x # element of Fp
+    ## XXX: smarter way for this?
+    let r = Fr[C].fromBig(rx.toBig())
```
Vindaar, line 197: We need to convert the coordinate in `Fp` to a scalar in `Fr`. The "best" way I could come up with is this to big / from big conversion. Is there a cleaner way?

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+  var h {.noinit.}: sha256
+  h.init()
+  h.update(message)
+  h.finish(result)
+
+proc toBytes(x: Fr[C] | Fp[C]): array[32, byte] =
+  let bi = x.toBig()
+  discard result.marshal(bi, bigEndian)
+
+proc toDER*(r, s: Fr[C]): seq[byte] =
```
mratsim, line 39: There should be no allocations for signature algorithms.

Also DER should likely go in `constantine/serialization`

constantine/ecdsa/ecdsa.nim
mratsim: For now it's fine not to address this, AFAIK most protocols use RFC6979 to avoid adding the RNG to the attack surface. If the RNG fails to produce data, I'm not even sure what it means at the OS level, no more entropy maybe?

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+  let m = FF.getModulus()
+  var b: matchingBigInt(FF.Name)
+
+  while b.isZero().bool or (b > m).bool:
+    ## XXX: raise / what else to do if `sysrand` call fails?
+    doAssert b.limbs.sysrand()
+
+  result.fromBig(b)
+
+proc arrayWith[N: static int](val: byte): array[N, byte] =
```
mratsim, line 93:
```suggestion
proc arrayWith[N: static int](val: byte): array[N, byte] {.noInit.}=
```

Same remark on returning big values

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+  while true:
+    # Step h.1 (init T to zero) and h.2:
+    # `V = HMAC_K(V)`
+    # `T = T || V`
+    # We do not need to accumulate a `T`, because we use SHA256 as a hash
+    # function (256 bits) and Secp256k1 as a curve (also 256 big int).
+    hmac.round(k, v, v) # v becomes T
+
+    # Step h.3: `k = bits2int(T)`
+    var candidate: Fr[C]
```
mratsim, line 153:
```suggestion
    var candidate {.noinit.}: Fr[C]
```

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+      return candidate
+
+    # Step h.3 failure state:
+    # `K = HMAC_K(V || 0x00)`
+    # `V = HMAC_K(V)`
+    # Try again if invalid
+    hmac.round(k, k, v, [byte 0x00])
+    hmac.round(k, v, v)
+
+proc generateNonce(kind: NonceSampler, msgHash, privateKey: Fr[C]): Fr[C] =
```
mratsim, line 168:
```suggestion
proc generateNonce(kind: NonceSampler, msgHash, privateKey: Fr[C]): Fr[C] {.noinit.} =
```

constantine/ecdsa/ecdsa.nim
mratsim: This is the best generic conversion.

In theory for secp256k1, you could take `mod r` because Fp/Fr can be both use canonical representation and still have a fast reduction (#445) however at the moment only Fp uses canonical repr and fast reduction while Fr uses Montgomery representation and Montgomery reduction.

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+
+    if bool(r.isZero()):
+      continue # try again
+
+    # Calculate s
+    # `s = (k⁻¹ · (h + r · p)) (mod n)`
+    # with `h`: message hash as `Fr[C]` (if we didn't use SHA256 w/ 32 byte output
+    # we'd need to truncate to N bits for N being bits in modulo `n`)
+    k.inv()
+    var s = (k * (message_hash + r * privateKey))
```
mratsim, line 207: It might be faster to do k*privateKey and then a multiscalarmul([k, k*privkey], [message_hash, r]).

Note that the out-of-place functions are for convenience when debugging/testing/developping but not for production code (https://github.com/mratsim/constantine/pull/413)

mratsim (commented):
Note that for Ethereum the hash function is Keccak. But it's good to have both and be hash agnostic to help test.

The out-of-place function will need to be replaced by in-place.

You can check `constantine/signatures/bls_signatures.nim` in general for API guidance.
Also https://github.com/bitcoin-core/secp256k1 which is the reference implementation for Bitcoin and Ethereum to implement serializer without dynamic allocation.
And https://github.com/status-im/nim-eth/blob/master/eth/common/keys.nim for how secp256k1 is actually used in Ethereum

The PEM/DER part is unused in blockchain, but it's helpful for testing against OpenSSL, and if we ever want to add TLS/QUIC support to Constantine, it wil be necessary.

Regarding testing vs openssl if the API for ECDSA is less messy than hashes, it can be used as a library like so https://github.com/mratsim/constantine/blob/13d5bb73f382ea95304962cabd0d159313472801/tests/t_hash_sha256_vs_openssl.nim#L8-L54 and we can remove the random messages json

constantine/ecdsa/ecdsa.nim
Vindaar: Not sure either to be honest, but given that we _do_ return a `bool` I didn't just want to discard it.

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+  ## a simple uniform random sampler. Alternatively a deterministic
+  ## sampler based on message hash and private key.
+  NonceSampler* = enum
+    nsRandom, ## pure uniform random sampling
+    nsRfc6979 ## deterministic according to RFC 6979
+
+# For easier readibility, define the curve and generator
+# as globals in this file
+const C* = Secp256k1
+const G = Secp256k1.getGenerator("G1")
```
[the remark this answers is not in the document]
Vindaar: Yup, fair enough. I wrote it by defining the curve like this so that the code itself is already an (almost) valid generic / static proc. Didn't know if we want to make it fully generic immediately or start with secp256k1 only. Will address it.

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+
+proc hashMessage(message: string): array[32, byte] =
+  # Hash a given message
+  var h {.noinit.}: sha256
+  h.init()
+  h.update(message)
+  h.finish(result)
+
+proc toBytes(x: Fr[C] | Fp[C]): array[32, byte] =
```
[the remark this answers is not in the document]
Vindaar: Ah, that either slipped my mind or I just wasn't aware of it! Good to know. Will change it to an in-place variant.

constantine/ecdsa/ecdsa.nim
Vindaar: Well, `toDER` is mainly for testing / convenience as also mentioned in its docstring. That's why I went with the simple `seq` approach here (mainly because we cannot determine at compile time what length the resulting DER encoded data has (due to prefix `0` byte or dropped leading zero bytes). Otherwise I assume one needs to always use an array that can fit the longest variant and truncate later.

constantine/ecdsa/ecdsa.nim
```diff
@@ -0,0 +1,356 @@
...
+  let R = point1 + point2
+
+  # 5. Get x coordinate and convert to Fr (like in signing)
+  let x = R.getAffine().x
+  let r_computed = Fr[C].fromBig(x.toBig())
+
+  # 6. Verify r_computed equals provided r
+  result = bool(r_computed == signature.r)
+
+proc generatePrivateKey*(): Fr[C] =
```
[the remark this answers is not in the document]
Vindaar: Yeah, I assumed having _a_ way to generate a random private key might be useful. For testing we can just move it to a test file of course.

constantine/ecdsa/ecdsa.nim
Vindaar: I understand. Still hasn't fully sunk in (intuitively) to avoid out-of-place procs! My bad. :grin:

constantine/ecdsa/ecdsa.nim
mratsim: I would defer then to how libsecp256k1 (https://github.com/bitcoin-core/secp256k1) or BearSSL (https://bearssl.org/#download-and-installation) do it

libsecp256k1 is blockchain-focused so some TLS oriented serialization is not available.
Neither allocates.

But it's fine to put a note for now with TODO: zero-alloc

Vindaar: Changed the implementation to something inspired by Bitcoin's implementation. Now have a `DERSignature` type  which has a 72 byte array buffer & an actual length and fill up to `len` bytes.

(We could of course also just have the proc take `sig: var array[72, byte], outlen: var int` parameters instead)

Vindaar:
Just pushed some more changes for most remaining comments.

Still a few things to do. Essentially put serialization into proper submodule, update test cases to use library of OpenSSL and adjust API for specific ECDSA over secp256k1 impl (the latter will be done as part of the test rewrite).

- [X] Make API generic in `signatures` and add `ecdsa_secp256k1.nim`
  in `constantine/`
  - [X] Use API for verification `coreVerify`
  - [X] Use API for hashes
  - [X] Make hash agnostic
    - [X] Signing, done
    - [X] Verification
  - [X] Make all the code generic under
    - [X] Different hash functions
      - [X] Digest arrays, use `.digestSize` of the hash function
    - [X] Different curves
      - [X] Generator
      - [X] BigInt
    - [X] Handle different sizes of hash functions & curve orders,
      such that we truncate the hash digest in the signing and
      verification process. 
- [X] Move some common signature operations to their own file `signatures/common_signature_ops.nim`      
- [ ] Adjust API for ECDSA secp256k1 
- [ ] Move serialization logic to `serializtion.nim`
  - [ ] Submodule for PEM files specific to secp256k1 for now
  - [ ] Submodule for DER signatures general to ECDSA
- [ ] OpenSSL wrapper for tests
- [ ] Make serialization to PEM files generic under different curves
  (low priority)

Vindaar:
Mostly done now, I'd say. Addressed everything I could think of. The public API for ECDSA over secp256k1 could be extended a bit more, I suppose.

I wanted to use OpenSSL for PEM file writing as part of the tests to verify our serializers, but ended up giving up after wasting 2 hours or so with either getting segfaults in OpenSSL procedures or only having empty PEM files. So instead for that we still call to an external process to generate PEM files from our private and public keys, sign and verify a message.

I'll rebase and fix the conflict of the two files later. I pulled out the OpenSSL wrapper into a single file. Guess the files were updated on master.

The two main features lacking are public key recovery and batch verification.

Vindaar:
Just rebased and force pushed the changes to fix the merge conflict.

mratsim:
From testing with OpenSSL for hash functions:

- On MacOS, OpenSSL isn't shipped it's LibreSSL instead so APIs can change, and they did because OpenSSL 3.0 was a big break.
- On Windows, I'm not sure if we're dealing with 1..x or LibreSSL or yet another fork

but it's fine to skip testing vs OpenSSL on non-linux platforms for now.

constantine/serialization/codecs_ecdsa_secp256k1.nim
```diff
@@ -0,0 +1,163 @@
...
+
+import
+  constantine/named/algebras,
+  constantine/platforms/primitives,
+  constantine/math/arithmetic/finite_fields,
+  constantine/math/elliptic/ec_shortweierstrass_affine,
+  constantine/math/io/io_bigints,
+  constantine/ecdsa_secp256k1
+
+import std / [strutils, base64, math, importutils]
```
mratsim, line 43: This needs a follow up issue to remove the dependencies

constantine/serialization/codecs_ecdsa_secp256k1.nim
```diff
@@ -0,0 +1,163 @@
...
+## it currently still contains stdlib dependencies. Most of those are trivial, with the
+## exception of a base64 encoder.
+## Having a ANS1.DER encoder (and maybe decoder in the future) for SEC1 private and
+## public keys would be nice to have in CTT, I think (at least for the curves that
+## we support for the related operations; secp256k1 at the moment).
+
+proc wrap(s: string, maxLineWidth = 64): string =
+  ## Wrap the given string at `maxLineWidth` over multiple lines
+  let lines = s.len.ceilDiv maxLineWidth
+  result = newStringOfCap(s.len + lines)
```
mratsim, line 123: In the refactoring issue, we should also remove the allocation.

ECDSA is used in embedded devices and in context where memory cannot be allocated dynamically like trusted enclaves or smart-cards, and I'd like constantine to be alloc-free for those use-cases

constantine/serialization/codecs_ecdsa_secp256k1.nim
```diff
@@ -0,0 +1,163 @@
...
+  for i in 0 ..< lines:
+    let frm = i * maxLineWidth
+    let to = min(s.len, (i+1) * maxLineWidth)
+    result.add s[frm ..< to]
+    if i < lines-1:
+      result.add "\n"
+
+proc toPemFile*(publicKey: PublicKey): string =
```
mratsim, line 131: This is not a file

constantine/serialization/codecs_ecdsa_secp256k1.nim
```diff
@@ -0,0 +1,163 @@
...
+  # 1. Convert public key to ASN.1 DER
+  var derB: array[88, byte]
+  privateAccess(PublicKey) # for `raw` access
+  derB.toPemPublicKey(publicKey.raw)
+  # 2. Encode bytes in base64
+  let der64 = derB.encode().wrap()
+  # 3. Wrap in begin/end public key template
+  result = "-----BEGIN PUBLIC KEY-----\n"
+  result.add der64 & "\n"
+  result.add "-----END PUBLIC KEY-----\n"
```
mratsim, line 145: It seems like the length of the result is fixed?
So in the refactoring issue for ECDSA serialization, we should replace the string output with a `var array` or `var openarray` (+ size check)

constantine/signatures/common_signature_ops.nim
```diff
@@ -0,0 +1,26 @@
...
+  ##
+  ## The secret key MUST be in range (0, curve order)
+  ## 0 is INVALID
+  const Group = Pubkey.G
+  type Field = Pubkey.F
+  const EC = Field.Name
+
+  var pk {.noInit.}: EC_ShortW_Jac[Field, Group]
+  pk.setGenerator()
+  pk.scalarMul(seckey)
```
mratsim, line 25: I'm pretty sure lattices don't derive public keys the same way ;).

Probably better to rename the file `ecc_sig_ops.nim` or something.

mratsim (commented):
Review part 1.

We will need a followup refactoring on serialization to remove stdlib dependencies and ideally allocation as well.

Part 2 next year.

constantine/serialization/codecs_ecdsa.nim
```diff
@@ -0,0 +1,161 @@
...
+  constantine/math/elliptic/[ec_shortweierstrass_affine],
+  constantine/platforms/[abstractions, views],
+  constantine/serialization/codecs # for fromHex and (in the future) base64 encoding
+
+type
+  ## Helper type for ASN.1 DER signatures to avoid allocation.
+  ## Has a `data` buffer of 72 bytes (maximum possible size for
+  ## a signature for `secp256k1`) and `len` of actually used data.
+  ## `data[0 ..< len]` is the actual signature.
+  DERSignature*[N: static int] = object
```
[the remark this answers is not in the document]
Vindaar: Haha, I think we can live with that. ;) We speak perfect German anyway, so no one would confuse 'DerSignature' with 'Der Signatur'. Only the latter would be wrong German after all!

constantine/serialization/codecs_ecdsa_secp256k1.nim
Vindaar: Yes, as also mentioned in the docstring at the top. I'll open an issue for it!

constantine/serialization/codecs_ecdsa_secp256k1.nim
Vindaar: Yes, the length is fixed (in contrast to DER encoded signatures). The `newStringOfCap` was partially out of laziness as the inclusion of the stdlib `base64` encoder renders the code anyhow in need of an overhaul. Will add as part of the issue I'll open.

constantine/serialization/codecs_ecdsa_secp256k1.nim
Vindaar: Issue: https://github.com/mratsim/constantine/issues/506

constantine/signatures/ecdsa.nim
```diff
@@ -0,0 +1,311 @@
...
+  let m = FF.getModulus()
+  var b: matchingBigInt(FF.Name)
+
+  while b.isZero().bool or (b > m).bool:
+    ## XXX: raise / what else to do if `sysrand` call fails?
+    doAssert b.limbs.sysrand()
+
+  result.fromBig(b)
+
+proc arrayWith[N: static int](res: var array[N, byte], val: byte) =
```
mratsim, line 86:
```suggestion
proc byteArrayWith(N: static int, val: byte): array[N, byte] {.noInit, inline.} =
```

mratsim: This is only used twice and just after initialization. We can avoid bad codegen with `{.inline.}` and we can avoid avoid an extra `memset(0)` with `{.noInit.}`

constantine/signatures/ecdsa.nim
```diff
@@ -0,0 +1,311 @@
...
+  # `r` bytes to a BigInt, reduces modulo prime order `p` and converts to
+  # a byte array).
+  var privKeyBytes {.noinit.}: array[N, byte]
+  privKeyBytes.toBytes(privateKey)
+
+  # Initial values
+  # Step b: `V = 0x01 0x01 0x01 ... 0x01`
+  var v: array[N, byte]; v.arrayWith(byte 0x01)
+  # Step c: `K = 0x00 0x00 0x00 ... 0x00`
+  var k: array[N, byte]; k.arrayWith(byte 0x00)
```
mratsim, line 131: awkward, see change to `arrayWith` to do declaration and init.

constantine/signatures/ecdsa.nim
```diff
@@ -0,0 +1,311 @@
...
+  # 4. Compute u₁G + u₂Q
+  var
+    point1 {.noinit.}: EC_ShortW_Jac[Fp[Name], G1]
+    point2 {.noinit.}: EC_ShortW_Jac[Fp[Name], G1]
+  # Generator of the curve
+  const G = publicKey.F.Name.getGenerator($publicKey.G)
+  point1.scalarMul(u1, G)
+  point2.scalarMul(u2, publicKey)
+  var R {.noinit.}: EC_ShortW_Jac[Fp[Name], G1]
+  R.sum(point1, point2)
```
mratsim, line 292: As mentioned in https://github.com/mratsim/constantine/pull/504#discussion_r1900762614 this should be a dedicated proc, which would be subject to optimization in a later PR as it's a bottleneck for blockchain (signatures are verified all the time) and TLS.

mratsim (commented):
Review part 2. LGTM

constantine/signatures/ecdsa.nim
Vindaar: Yup. I'll open an issue about these two cases (here and the other PR).

constantine/ecdsa_secp256k1.nim
```diff
@@ -0,0 +1,68 @@
...
+  bool(a.r == b.r and a.s == b.s)
+
+proc sign*(sig: var Signature,
+           secretKey: SecretKey,
+           message: openArray[byte],
+           nonceSampler: NonceSampler = nsRandom) {.libPrefix: prefix_ffi, genCharAPI.} =
+  ## Sign `message` using `secretKey` and store the signature in `sig`. The nonce
+  ## will either be randomly sampled `nsRandom` or deterministically calculated according
+  ## to RFC6979 (`nsRfc6979`)
+  sig.coreSign(secretKey.raw, message, sha256, nonceSampler)
```
mratsim, line 54: I think it's Bitcoin protocol that uses SHA256?

We can rename the file btc_ecdsa_secp256k1.

mratsim (commented):
LGTM, merging

PR #494, opened 2024-12-22
Keccak256 and SHA3-256

mratsim:
This PR adds support for SHA3-256 and Keccak256, both being based on the Keccak permutation and differing only by single byte in the padding rule.

Benchmarks without optimization vs Nim or C implementations
image

Asymptotically we're 94% the speed of OpenSSL without assembly or vectorization.
For small inputs like 32B we're almost 2x faster.

This is quite important for Ethereum as Ethereum Patricial-Merkle trie is built using Keccak.

There is still significant room for optimization as while we're faster than RustCrypto for large inputs, they are faster for small ones:
image

In fact, for Constantine, OpenSSL and Keccak tiny, the speed is the same for 128 bytes input and below.
This is probably because they all use an intermediate buffer of size 136 (keccak "rate").


mratsim:
Actually the Rust benchmarks are misleading, RustCrypto only benches "hash.update()" and not the full hash, here is a corrected bench:

image

see https://github.com/mratsim/constantine/issues/495#issuecomment-2561915443

PR #500, opened 2024-12-26
Improve Rust build script

DaniPopes:
- Report `nimble` build errors
  - I encountered a few build errors due to missing `llvm-as`, which was not reported on the console.
- Improve rustc-link-lib directive by letting the compiler find the correct target-specific name
  - I believe this should work, will revert if CI fails
- Reduce rerun-if-changed directives to reduce unnecessary re-compilations
  - Currently every single `cargo build` rebuilds constantine which takes like 15s.


mratsim:
Thank you!

Issue #36, opened 2020-06-04
Strauss-Shamir trick: [a]P + [b]Q


mratsim:
It is often needed to compute scalar multiplication in parallel via
```
R <- [a]P + [b]Q
```

This is usually known as the Strauss-Shamir trick, see https://eprint.iacr.org/2003/257.pdf or https://stackoverflow.com/questions/50993471/ec-scalar-multiplication-with-strauss-shamir-method

![image](https://user-images.githubusercontent.com/22738317/83814717-1afd6580-a6bf-11ea-99f2-81ef8e159f4c.png)

![image](https://user-images.githubusercontent.com/22738317/83814454-f1dcd500-a6be-11ea-9373-b034d14d93c5.png)

Furthermore to reduce the number of point addition, the points P and Q can be recoded under the joint-sparse-form to minimize the total hamming weight.

paulmillr:
See https://gist.github.com/paulmillr/178042240169f0f531f8cc95e532f9db

It's slower if you use G precomputes, unfortunately.

mratsim:
For ease of simplicity, we can start by re-using the algorithms used to add split scalars resulting from endomorphism acceleration. This should provide a decent baseline since the doublings are shared.

PR #504, opened 2024-12-30
Add `ECRecover` EVM precompile

Vindaar:
This builds on top of the ECDSA implementation in PR #490 and adds public key recovery. Based on that implementing the `ECRecover` precompile is quite straightforward (recover an Ethereum address based on a given message digest and signature). 

We base our implementation of the precompile on Geth's implementation here:

https://github.com/ethereum/go-ethereum/blob/341647f1865dab437a690dc1424ba71495de2dd8/core/vm/contracts.go#L243-L272

and to a lesser extent appendix F in the Ethereum Yellow Paper.

For the general ECDSA public key recovery API I went with a variable time implementation. To my understanding this is essentially needed regardless, due to the fact that the signing process converts the coordinate of the point `R` in `Fp` to a scalar in `Fr`. While unlikely for Secp256k1 (because the subgroup order and the curve prime are both 256 bit integers), if the x coordinate is larger than the subgroup order, we reduce it. This reduction needs to be reversed in the recovery process requiring a loop.
In addition we have an `evenY` boolean argument, which decides if we recover the public key associated with the `R` coordinate with an even `y` coordinate or an odd one (if `y` is even then `-y` is odd in a finite field of odd size).

(I'll fix the CI failure in the other PR and will rebase this one on top of the other once that is done)

### DIFF
--- a/constantine.nimble
+++ b/constantine.nimble
@@ -609,7 +609,8 @@ const testDesc: seq[tuple[path: string, useGMP: bool]] = @[
   ("tests/t_ethereum_verkle_ipa_primitives.nim", false),
 
   # Signatures
-  ("tests/ecdsa/t_ecdsa_verify_openssl.nim", false),
+  # NOTE: Requires OpenSSL version >=v3.3 for to Keccak256 support
+  # ("tests/ecdsa/t_ecdsa_verify_openssl.nim", false),
 
   # Proof systems
   # ----------------------------------------------------------
--- a/constantine/ecdsa_secp256k1.nim
+++ b/constantine/ecdsa_secp256k1.nim
@@ -81,3 +81,21 @@ proc recoverPubkey*(
   ## `evenY == true` returns the public key corresponding to the
   ## even `y` coordinate of the `R` point.
   publicKey.raw.recoverPubkey(signature, message, evenY, H)
+
+proc recoverPubkey*(
+    publicKey: var PublicKey,
+    msgHash: Fr[Secp256k1],
+    signature: Signature,
+    evenY: bool
+) {.libPrefix: prefix_ffi.} =
+  ## Verify `signature` using `publicKey` for the given message digest
+  ## given as a scalar in the field `Fr[Secp256k1]`.
+  ##
+  ## `evenY == true` returns the public key corresponding to the
+  ## even `y` coordinate of the `R` point.
+  ##
+  ## As this overload works directly with a message hash as a scalar,
+  ## it requires no hash function. Internally, it also calls the
+  ## `verify` implementation, which already takes a scalar and thus
+  ## requires no hash function there either.
+  publicKey.raw.recoverPubkeyImpl_vartime(signature, msgHash, evenY)
--- a/constantine/ecdsa_secp256k1.nim
+++ b/constantine/ecdsa_secp256k1.nim
@@ -66,3 +66,12 @@ func derive_pubkey*(public_key: var PublicKey, secret_key: SecretKey) {.libPrefi
   ##
   ## The secret_key MUST be validated
   public_key.raw.derivePubkey(secret_key.raw)
+
+proc recoverPubkey*(
+    publicKey: var PublicKey,
+    message: openArray[byte],
+    signature: Signature,
+    evenY: bool
+) {.libPrefix: prefix_ffi, genCharAPI.} =
+  ## Verify `signature` using `publicKey` for `message`.
+  publicKey.raw.recoverPubkey(signature, message, evenY, sha256)
--- a/constantine/ecdsa_secp256k1.nim
+++ b/constantine/ecdsa_secp256k1.nim
@@ -9,7 +9,7 @@
 import
   constantine/zoo_exports,
   constantine/signatures/ecdsa,
-  constantine/hashes/h_sha256,
+  constantine/hashes,
   constantine/named/algebras,
   constantine/math/elliptic/[ec_shortweierstrass_affine],
   constantine/math/[arithmetic, ec_shortweierstrass],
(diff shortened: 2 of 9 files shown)


constantine/signatures/ecdsa.nim
```diff
@@ -308,4 +301,110 @@ func coreVerify*[Pubkey, Sig](
...
+    var u1 {.noinit.}, u2 {.noinit.}: Fr[Name]
+    u1.prod(msgHash, rInv)     # `u₁ = m·r⁻¹`
+    u1.neg()                   # `u₁ = -m·r⁻¹`
+    u2.prod(signature.s, rInv) # `u₂ = s·r⁻¹`
+
+    var Q {.noinit.}: ECJac # the potential public key
+    var point1 {.noinit.}, point2 {.noinit.}: ECJac
+    point1.scalarMul(u1, G)    # `p₁ = u₁ * G`
+    point2.scalarMul(u2, R)    # `p₂ = u₂ * R`
+    Q.sum(point1, point2)      # `Q = p₁ + p₂`
```
mratsim, line 375: This comes up quite often for ECDSA and should be a dedicated proc, unsure about the name yet `dualScalarMul`, `simul`, `lincomb`.

The multiscalarmul_vartime proc has too much overhead for just 2:
image
but there are the Strauss or the Yao algorithms to accelerate this, especially given that it's a bottleneck for TLS:
- https://github.com/mratsim/constantine/issues/36
- https://cr.yp.to/papers/pippenger-20020118-retypeset20220327.pdf

mratsim: Also the `scalarMulEndo` routines like this https://github.com/mratsim/constantine/blob/6b65b0e77a4ea3435b7ba2a7a9e01aaf704653a5/constantine/math/elliptic/ec_scalar_mul_vartime.nim#L252-L331

actually are implemented in 2 steps:
- splitting a scalar in 2 or 4
- and computing a dual or quadruple scalar multiplications efficiently, including in constant-time.

So it should be possible to isolate those internals in a separate PR and redirect the publicly exposed `dualScalarMul` to use these and minimize extra code that needs to be written.

Vindaar: Issue for it here: https://github.com/mratsim/constantine/issues/507

constantine/ecdsa_secp256k1.nim
```diff
@@ -47,22 +47,55 @@ func signatures_are_equal*(a, b: Signature): bool {.libPrefix: prefix_ffi.} =
 proc sign*(sig: var Signature,
            secretKey: SecretKey,
            message: openArray[byte],
-           nonceSampler: NonceSampler = nsRandom) {.libPrefix: prefix_ffi, genCharAPI.} =
+           nonceSampler: NonceSampler = nsRandom,
+           H: type CryptoHash = sha256) {.libPrefix: prefix_ffi, genCharAPI.} =
```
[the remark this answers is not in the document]
Vindaar: Oh, right! Yeah, let's go with that. It's trivial to add another at some point in the future after all.

constantine/ethereum_evm_precompiles.nim
```diff
@@ -48,6 +50,8 @@ type
     cttEVM_PointNotOnCurve
     cttEVM_PointNotInSubgroup
     cttEVM_VerificationFailure
+    cttEVM_InvalidSignature
+    cttEVM_InvalidV # `v` of signature `(r, s, v)` is invalid
```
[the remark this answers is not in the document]
Vindaar: In practice, I'd say no. Here I was just a bit hasty in mapping the errors from the test vectors to enum elements. Will go with `MalformedSignature` as below for both.

constantine/signatures/ecdsa.nim
```diff
@@ -308,4 +301,110 @@ func coreVerify*[Pubkey, Sig](
...
+  # 1. Set to neutral so if we don't find a valid signature, return neutral
+  recovered.setNeutral()
+  const G = Name.getGenerator($G1)
+
+  let rInit = signature.r.toBig() # initial `r`
+  var x1 = Fp[Name].fromBig(signature.r.toBig()) # as coordinate in Fp
+  let M = Fp[Name].fromBig(Fr[Name].getModulus())
+
+  # Due to the conversion of the `x` coordinate in `Fp` of the point `R` in the signing process
+  # to a scalar in `Fr`, we potentially reduce it modulo the subgroup order (if `x > M` with
```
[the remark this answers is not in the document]
Vindaar: Good catch on the missing `=` case, heh.

Vindaar:
Rebased onto the ECDSA PR branch & made the ECDSA file for secp256k1 specific to Ethereum now. The CI should pass now (aside from the maybe failing Windows test case like in the other where it gets terminated at some point?).

Vindaar:
CI is failing also on Linux, because of an outdated OpenSSL version on `ubuntu-latest` (= Ubuntu 22.04).

Ubuntu 22.04 ships OpenSSL version 3.0.2: 
https://packages.ubuntu.com/jammy/openssl

while support for Keccak256 was only added in version 3.2 according to this issue comment in July 2024:
https://github.com/openssl/openssl/issues/19304#issuecomment-2225962138

"Funnily" enough, Keccak is not mentioned anywhere in the actual changelog: 
https://github.com/openssl/openssl/blob/openssl-3.2/CHANGES.md

constantine.nimble
```diff
@@ -607,6 +607,9 @@ const testDesc: seq[tuple[path: string, useGMP: bool]] = @[
   ("tests/t_ethereum_verkle_primitives.nim", false),
   ("tests/t_ethereum_verkle_ipa_primitives.nim", false),
 
+  # Signatures
+  ("tests/ecdsa/t_ecdsa_verify_openssl.nim", false),
```
mratsim, line 611: This can be commented out until Ubuntu 25.04 is available in CI (probably June)

mratsim (approved):
LGTM, the PR needs a rebase

constantine.nimble
Vindaar: Are the non LTS releases ever part of the Github Actions CI images?

Vindaar:
Rebased, force pushed and taken out the ECDSA test for the time being.

mratsim (approved):

PR #529, opened 2025-02-15
remove `when nimvm` condition to import

metagn:
Both branches of a `when nimvm` statement are compiled, so this import is always processed.

### DIFF
--- a/constantine/math/arithmetic/finite_fields.nim
+++ b/constantine/math/arithmetic/finite_fields.nim
@@ -39,10 +39,7 @@ when UseASM_X86_64:
 when UseASM_ARM64:
   import ./assembly/limbs_asm_modular_arm64
 
-when nimvm:
-  from constantine/named/deriv/precompute import montyResidue_precompute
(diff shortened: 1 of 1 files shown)


mratsim:
So the current code is correct no?

At compile-time I use the code loaded by the `when nimvm: import` and at runtime I use normal code.

metagn:
In this case (the import) there is no difference between using `when nimvm` or writing out either branch. Import statements are not processed during code generation but in the typechecking phase, both branches of `when nimvm` are always typechecked. This also means any symbols defined/imported in either `when nimvm` branch are always defined/imported.

This doesn't mean that these symbols are included in code generation, if procs are only used in contexts that generate code for the VM and not the backend, then dead code elimination will prevent them from being generated. This wouldn't include things like globals as dead code elimination does not apply to them, but in this case there doesn't seem to be any globals. In any case the `{.compileTime.}` annotation exists for both routines and globals to prevent their generation despite its issues.

PR #554, opened 2025-08-02
fix NVRTC CUDA backend due to incompatible pointer types

Vindaar:
Our previous approach of just a bare `distinct pointer` now leads to a mismatched pointer error on modern GCC / Clang.

Similarly for `cuGetErrorString` we now *need* a `const char**` argument and not just a `char **` argument...

Edit:

And a couple of other things I just noticed running some of the NVRTC tests locally:
- I hadn't ever committed the correct `slct` proc syntax (takes `int32` instead of `uint32`)
- fix a regression where we would generate duplicate device pointers if one passes in the same argument as an input twice
- avoids the need to hardcode the limb size in `defCoreFieldOps` 

Note: I still see test failures here that elude me at the moment. I'll need some time to debug them. Still, at least the code compiles and runs again, heh.

### DIFF
--- a/constantine/platforms/abis/nvidia_abi.nim
+++ b/constantine/platforms/abis/nvidia_abi.nim
@@ -32,6 +32,14 @@ const libCuda = "(libcuda.so|libcuda.so)"
 #
 # We need to use the lower-level driver API for JIT modules loading and reloading
 
+
+## Sigh. Due to the `incompatible-pointer-types` now standard in GCC / Clang, we need to
+## define a real `const char*` type for e.g. `cuGetErrorString`. Otherwise we get errors.
+type
+  cstringConstImpl* {.importc: "const char*".} = cstring
+  constChar* = distinct cstringConstImpl
+
+
 type
   CUresult* {.size: sizeof(cint).} = enum
     ##  The API call returned with no errors. In the case of query calls, this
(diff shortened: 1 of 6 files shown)


Vindaar:
Added a fix for the issues I was encountering running the basic test cases. Essentially a change I made (defaulting to `passStructByPointer = false`) broke the simple test cases, in which we pass in a `BigInt` "by value" (in terms of function signature on the CUDA kernel side). 
See commit https://github.com/mratsim/constantine/pull/554/commits/370c117eedbda1df98a0aed3fcc91cba28f39c36 for a bit more about it.

In any case, the real fix will need to wait. We _can_ make it so that this is all done correctly for the caller, but that needs not just information about the data we pass _into_ the kernel, but also about the signature of the kernel itself (which we currently don't have information about in the context of the calling code).

I have code for a WebGPU / WGSL backend, which contains changes that make this somewhat easier. So I'd say we merge this and then in an upcoming PR that introduces the new backend, I'll provide a proper fix.

constantine/platforms/abis/nvidia_abi.nim
```diff
@@ -477,14 +485,27 @@ type
...
-  CUstream* = distinct pointer
-  CUdeviceptr* = distinct pointer
 
-  CUlinkState_st = object
-  CUlinkState* = ptr CUlinkState_st
+  CUcontext* {.importc: "CUcontext", header: "cuda.h".} = distinct pointer
+  CUmodule*  {.importc: "CUmodule", header: "cuda.h".} = distinct pointer
+  CUfunction* {.importc: "CUfunction", header: "cuda.h".} = distinct pointer
+  CUstream* {.importc: "CUstream", header: "cuda.h".} = distinct pointer
+  CUlinkState* {.importc: "CUlinkState", header: "cuda.h".} = distinct pointer # ptr CUlinkState_st
```
mratsim, line 493: Do we pass the location of `cuda.h` somewhere or do we assume it's generally in the path.

But in that case shouldn't we use `header: "<cuda.h>"` ?

My worry with this is Windows support, working with headers/dependencies there is just hell and was fuel to rip out BLAS dependencies from Arraymancer: https://github.com/mratsim/Arraymancer/issues/422

Vindaar: I don't like the dependency either, but we already had it: https://github.com/mratsim/constantine/blob/master/constantine/platforms/abis/nvidia_abi.nim#L831

(I had to add this due to the issues described in https://github.com/SciNim/nimcuda/issues/27, part of the initial NVRTC PR)

PR #565, opened 2025-08-21
WebGPU GPU compiler backend improvements (and small CUDA)

Vindaar:
This improves the initial WebGPU backend added in #564 to make it practically useful. Aside from adding a lot of features and bug fixes for the WGSL spec, it also makes changes towards making the entire `cuda` macro more useful:

- Nim generic functions can now be used inside the GPU code. We emit one function for each generic instantiation. This of course includes e.g. `static int` arguments. We now _could_ replace the templates that receive a static parameter by a regular generic, e.g. for the `BigInt` type and generate the correct functions and types based on what is actually instantiated. For the moment we don't update the code in `gpu_field_ops.nim` though.
- Functions don't _need_ to be defined inside the `cuda` block anymore. We will pull any code used inside of the GPU block in, if it is called. This is a good step towards making code naturally run on CPU and GPU for example. Only the code called from the host still needs to be in the `cuda` macro. Note that keeping GPU code in `templates` can still be a good idea to not pollute the namespace and if one wants a clearer separation between GPU and CPU code.
- Similarly types are also pulled in from outside the macro, if they are used.

As a result also the CUDA backend was updated slightly to follow a similar structure as the WGSL backend. That is have a `preprocess` pass before the actual codegen pass to scan the global functions for functions and types actually used.

### DIFF
--- a/constantine/math_compiler/experimental/backends/cuda.nim
+++ b/constantine/math_compiler/experimental/backends/cuda.nim
@@ -43,6 +43,7 @@ proc gpuTypeToString*(t: GpuTypeKind): string =
   of gtVoidPtr: "void*"
   of gtObject: "struct"
   of gtString: "const char*"
+  of gtUA: "" # `UncheckedArray` by itself is nothing in CUDA
   else:
     raiseAssert "Invalid type : " & $t
 
--- a/constantine/math_compiler/experimental/backends/cuda.nim
+++ b/constantine/math_compiler/experimental/backends/cuda.nim
@@ -6,7 +6,7 @@
 #   * Apache v2 license (license terms in the root directory or at http://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import std / [macros, strformat, strutils, sugar, sequtils]
+import std / [macros, strformat, strutils, sugar, sequtils, tables, algorithm]
 
 import ../gpu_types
 import ./common_utils
@@ -119,10 +119,61 @@ proc genFunctionType*(typ: GpuType, fn: string, fnArgs: string): string =
 proc genMemcpy(lhs, rhs, size: string): string =
   result = &"memcpy({lhs}, {rhs}, {size})"
 
+proc scanFunctions(ctx: var GpuContext, n: GpuAst) =
+  ## Iterates over the given function and checks for all `gpuCall` nodes. Any function
+  ## called in the scope is added to `fnTab`. This is a form of dead code elimination.
+  case n.kind
+  of gpuCall:
+    let fn = n.cName
+    if fn in ctx.allFnTab:
+      # Check if any of the parameters are pointers (otherwise non generic)
+      if fn notin ctx.fnTab: # function not known, add to `fnTab` (i.e. avoid code elimination)
+        let fnCalled = ctx.allFnTab[fn]
+        ctx.fnTab[fn] = fnCalled
+        # still "scan for functions", i.e. fill `fnTab` from inner calls
+        for ch in fnCalled:
+          ctx.scanFunctions(ch)
+      # else we don't do anything for this function
+    # Harvest functions from arguments to this call!
+    for arg in n.cArgs:
+      ctx.scanFunctions(arg)
+  else:
+    for ch in n:
+      ctx.scanFunctions(ch)
 
 proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string
 proc size(ctx: var GpuContext, a: GpuAst): string = size(ctx.genCuda(a))
 proc address(ctx: var GpuContext, a: GpuAst): string = address(ctx.genCuda(a))
+
+proc preprocess*(ctx: var GpuContext, ast: GpuAst, kernel: string = "") =
+
+  # 1. Add all data from `genericInsts` and `types` tables
+  #    In CUDA the types have to be before any possible global variables using
+  #    them!
+  for k, v in pairs(ctx.genericInsts):
+    ctx.allFnTab[k] = v
+  # And all the known types
+  for k, typ in pairs(ctx.types):
+    ctx.globalBlocks.add typ
+
+  # 2. Fill table with all *global* functions or *only* the specific `kernel`
+  #    if any given
+  var varBlock = GpuAst(kind: gpuBlock)
+  var typBlock = GpuAst(kind: gpuBlock)
+  ctx.farmTopLevel(ast, kernel, varBlock, typBlock)
+  ctx.globalBlocks.add varBlock
+  ctx.globalBlocks.add typBlock
+  ## XXX: `typBlock` should now always be empty, as we pass all
+  ## found types into `ctx.types`
+
+  # 3. Using all global functions, we traverse their AST for any `gpuCall` node. We inspect
+  #    the functions called and record them in `fnTab`.
+  let fns = toSeq(ctx.fnTab.pairs)
+  for (fnIdent, fn) in fns: # everything in `fnTab` at this point is a global function
+    # Get the original arguments (before lifting them) of this function. Needed in scan
+    # to check if `gpuCall` argument is a parameter.
+    let fnOrig = ctx.allFnTab[fnIdent]
+    ctx.scanFunctions(fn)
 
 proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
   ## The actual CUDA code generator.
@@ -143,10 +194,13 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
 
     # extern "C" is needed to avoid name mangling
     result = indentStr & "extern \"C\" " & attrs.join(" ") & " " &
-             fnSig & "{\n"
-
-    result &= ctx.genCuda(ast.pBody, indent + 1)
-    result &= "\n" & indentStr & "}"
+             fnSig
+    if ast.forwardDeclare:
+      result.add ";"
+    else:
+      result.add "{\n"
+      result &= ctx.genCuda(ast.pBody, indent + 1)
+      result &= "\n" & indentStr & "}"
 
   of gpuBlock:
     result = ""
@@ -296,3 +350,19 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
     echo "Unhandled node kind in genCuda: ", ast.kind
     raiseAssert "Unhandled node kind in genCuda: " & ast.repr
     result = ""
+
+proc codegen*(ctx: var GpuContext): string =
+  ## Generate the actual code for all pieces of the puzzle
+  # 1. generate code for the global blocks (types, global vars etc)
+  for blk in ctx.globalBlocks:
+    result.add ctx.genCuda(blk) & ";\n\n"
+
+  # 2. generate all regular functions
+  let fns = toSeq(ctx.fnTab.pairs)
+  for (fnIdent, fn) in fns:
+    let fnC = fn.clone()
+    fnC.forwardDeclare = true
+    result.add ctx.genCuda(fnC) & "\n"
+
+  for fnIdent, fn in ctx.fnTab:
+    result.add ctx.genCuda(fn) & "\n\n"
--- a/constantine/math_compiler/experimental/backends/cuda.nim
+++ b/constantine/math_compiler/experimental/backends/cuda.nim
@@ -256,7 +256,7 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
     result = ast.pOp & ctx.genCuda(ast.pVal)
 
   of gpuTypeDef:
-    result = "struct " & ast.tName & "{\n"
+    result = "struct " & gpuTypeToString(ast.tTyp) & "{\n"
     for el in ast.tFields:
       result.add "  " & gpuTypeToString(el.typ, el.name) & ";\n"
     result.add "}"
--- a/constantine/math_compiler/experimental/backends/cuda.nim
+++ b/constantine/math_compiler/experimental/backends/cuda.nim
@@ -58,6 +58,10 @@ proc gpuTypeToString*(t: GpuType, ident: string = "", allowArrayToPtr = false,
   var skipIdent = false
   case t.kind
   of gtPtr:
+    var t = t # if `ptr UncheckedArray`, remove the `gtUA` layer. No meaning on CUDA
+    if t.to.kind == gtUA:
+      t.to = t.to.uaTo
+
     if t.to.kind == gtArray: # ptr to array type
       # need to pass `*` for the pointer into the identifier, i.e.
       # `state: var array[4, BigInt]`
--- a/constantine/math_compiler/experimental/backends/wgsl.nim
+++ b/constantine/math_compiler/experimental/backends/wgsl.nim
@@ -512,6 +512,31 @@ proc updateSymsInGlobals(ctx: var GpuContext, n: GpuAst) =
     for ch in n:
       ctx.updateSymsInGlobals(ch)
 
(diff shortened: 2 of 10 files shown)


Vindaar:
NOTE: I just rebased onto current master, which means this PR is now out of date to the `webgpu-backend` PR. Hence the conflicts and ginormous diff. 

Once the first PR is merged, I'll rebase onto master again to cut that down. Don't want to rebase the other PR needlessly now.

constantine/math_compiler/experimental/nim_to_gpu.nim
```diff
@@ -1104,6 +1104,10 @@ proc toGpuAst*(ctx: var GpuContext, node: NimNode): GpuAst =
       if result.iName == "_":
         result.iName = "tmp_" & $ctx.genSymCount
         inc ctx.genSymCount
+      elif result.iName.startsWith("tmpTuple_"): # will have a Nim gensym'd suffix, replace by custom counter
+        result.iName = "tmpTuple_" & $ctx.genSymCount
+        result.iSym = result.iName & "_" & node.signatureHash() # and update the iSym to not be based on Nim's value either
+        inc ctx.genSymCount
```
mratsim, line 1120: Does that help with debugging? Different namespace?

Also is signatureHash expensive? We don't want to slow down the compiler too much when we start compiling 10k+ lines of code to the GPU.

mratsim:
This needs a rebase following #576

constantine/math_compiler/experimental/nim_to_gpu.nim
Vindaar: Reason for this is that we want the generated code to be deterministic, if possible. Because that allows us to test for regressions in code generation plainly by comparing the generated code to expectation (because there's no free CI runners with GPUs, we cannot actually run the code easily).

I don't expect `signatureHash` to be particularly expensive. It just computes an MD5 hash in the Nim VM. In any case, the Nim macro GPU compiler is already much faster than the NVRTC compilation stage. So this is not a bottleneck.

We need `signatureHash`, because we need unique symbols in the context of function overloads (generics or not) so that we can identify each symbol uniquely based on the associated type signature.

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -114,11 +130,133 @@ proc genFunctionType*(typ: GpuType, fn: string, fnArgs: string): string =
...
+               else: t.gFields
+  result = GpuType(kind: gtInvalid)
+  for f in flds:
+    if f.name == field.ident():
+      return f.typ
+
+proc getType(ctx: var GpuContext, arg: GpuAst, typeOfIndex = true): GpuType =
+  ## Tries to determine the underlying type of the AST.
+  ##
+  ## If `typeOfIndex` is `true`, we return the type of the index we access. Otherwise
```
mratsim: This part is a bit confusingly worded.

Let's say we have `let foo = [float32 1, 2, 4, 8]`

and we do `ctx.getType(foo[1])`, do we get `float32` or `int`.
- In the first case, we get the type of the item.
- In the second case we get the type of the index, but in which scenario would we need that proc with those input, it's very unnatural?

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -114,11 +130,133 @@ proc genFunctionType*(typ: GpuType, fn: string, fnArgs: string): string =
...
+  ## we return the type of the array / pointer.
+  ##
+  ## NOTE: Do *not* rely on this for `mutable` or `implicit` fields of pointer types!
```
mratsim, line 179: We might want to auto-handle `HiddenDeref`

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -114,11 +130,133 @@ proc genFunctionType*(typ: GpuType, fn: string, fnArgs: string): string =
...
+  of gpuConv: arg.convTo
+  of gpuCast: arg.cTo # ident of the thing we cast
+  else:
+    raiseAssert "Not implemented to determine type from node: " & $arg
+
+proc makeCodeValid(ctx: var GpuContext, n: var GpuAst) =
+  ## Addresses other AST patterns that need to be rewritten on CUDA. Aspects
+  ## that are rewritten include:
+  ##
+  ## - `Index` of `Deref` of `Ident` needs to be rewritten to `Index` of `Ident` if the
```
mratsim, line 215: Concrete examples would help maintenance, debugging or generalizing in the future

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -138,10 +276,13 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
 
     # extern "C" is needed to avoid name mangling
     result = indentStr & "extern \"C\" " & attrs.join(" ") & " " &
-             fnSig & "{\n"
-
-    result &= ctx.genCuda(ast.pBody, indent + 1)
-    result &= "\n" & indentStr & "}"
+             fnSig
+    if ast.forwardDeclare:
+      result.add ";"
```
mratsim:
```suggestion
      result.add ';'
```

nit: this reduces allocation

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -138,10 +276,13 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
...
+    else:
+      result.add "{\n"
+      result &= ctx.genCuda(ast.pBody, indent + 1)
+      result &= "\n" & indentStr & "}"
```
mratsim:
```suggestion
      result &= "\n" & indentStr & '}'
```

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -157,7 +298,9 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
       result.add "\n" & indentStr & "} // " & ast.blockLabel & "\n"
 
   of gpuVar:
-    result = indentStr & ast.vAttributes.join(" ") & " " & gpuTypeToString(ast.vType, ast.vName.ident())
+    let attrs = if ast.vAttributes.len > 0: ast.vAttributes.join(" ") & " "
```
mratsim:
```suggestion
    let attrs = if ast.vAttributes.len > 0: ast.vAttributes.join(" ") & ' '
```

constantine/math_compiler/experimental/backends/cuda.nim
```diff
@@ -224,9 +367,12 @@ proc genCuda*(ctx: var GpuContext, ast: GpuAst, indent = 0): string =
       result = ctx.genCuda(expandedBody, indent)
 
   of gpuBinOp:
-    result = indentStr & "(" & ctx.genCuda(ast.bLeft) & " " &
-             ast.bOp & " " &
-             ctx.genCuda(ast.bRight) & ")"
+    ctx.withoutSemicolon:
```
mratsim, line 376: Seems like it was defined in an earlier PR but that's a weird name, does it do string interpolation without a semicolon at the end?

constantine/math_compiler/experimental/backends/wgsl.nim
```diff
@@ -495,17 +497,253 @@ proc injectAddressOf(ctx: var GpuContext, n: var GpuAst) =
...
+    for ch in n:
+      ctx.checkCodeValid(ch)
+
+proc pullConstantPragmaVars(ctx: var GpuContext, blk: var GpuAst) =
+  ## Filters out all `var foo {.constant.}: dtype` from the `globalBlocks` and adds them to
+  ## the `globals` of the context. Such variables are *not* regular global constants, but rather
+  ## `storage` buffers, which are filled before the kernel is executed.
+  ##
+  ## XXX: Document current not ideal behavior that one needs to be careful to pass data into
+  ## `wgsl.fakeExecute` precisely in the order in which the `var foo {.constant.}` are defined
```
mratsim, line 682: Is that something fixable?

constantine/math_compiler/experimental/gpu_compiler.nim
```diff
@@ -54,25 +59,25 @@ type
...
 ## because then the typed code would evaluate the values before we
 ## can work with it from the typed macro.
-let blockIdx* = NvBlockIdx()
-let blockDim* = NvBlockDim()
-let gridDim* = NvGridDim()
-let threadIdx* = NvThreadIdx()
+let blockIdx* {.builtin.} = NvBlockIdx()
+let blockDim* {.builtin.} = NvBlockDim()
+let gridDim* {.builtin.} = NvGridDim()
+let threadIdx* {.builtin.} = NvThreadIdx()
```
mratsim: maybe those backend specific builtins should be in a backend specific file

constantine/math_compiler/experimental/gpu_field_ops.nim
```diff
@@ -189,52 +189,58 @@ template defWGSLHelpers*(): untyped {.dirty.} =
...
     let result = temp + carry_flag
     # Carry out if: temp overflowed OR (temp + carry overflowed)
     carry_flag = select(0'u32, 1'u32, (temp < a) or (result < temp))
     return result
 
-  # Add with carry in only
   proc add_ci(a: uint32, b: uint32): uint32 {.device.} =
+    # Add with carry in only.
+    # NOTE: `carry_flag` is not reset, because the next call after
+    # an `add_ci` *must* be `add_co` or `sub_bo`, but never
```
mratsim, line 210: A better explanation is that  add_ci consumes the carry flag so next operation must be independent from it. i.e. `add` is fine if for completely separate operands/purposes

constantine/math_compiler/experimental/backends/cuda.nim
Vindaar: When encountering a `gpuBlock` in the code generation, if we are inside of a (nested) `withoutSemicolon` template, we do not emit a `;` at the end of statements in a block.

The idea is that we can have a `gpuBlock` as part of, for example, an if condition. We need a way to suppress the emission of a `;` at the end of each element of the condition.

`withoutSemicolon` simply changes the `skipSemicolon` field of the `GpuContext` until the end of the template. See the codegen part of `gpuBlock` above.

constantine/math_compiler/experimental/backends/wgsl.nim
Vindaar: Yes, it's fixable. Mostly I wasn't sure about the cleanest solution and given that `fakeExecute` is not even part of this PR at the moment, I left that for the future to save time.

constantine/math_compiler/experimental/gpu_compiler.nim
Vindaar: Yeah, probably a good idea.

constantine/math_compiler/experimental/gpu_field_ops.nim
Vindaar: My point of the comment was that resetting the carry flag explicitly would be costly and only an issue if the operations are used in the wrong order. 

Ideally we'd just set `carry_flag = 0` at the end of this proc.

constantine/math_compiler/experimental/backends/cuda.nim
Vindaar: It's a bit confusing, yes. I'll rephrase it.

By "type of the index", I mean it returns the type of an element of the `gpuIndex` operation (that is a `[]` accessor). The default is `typeOfIndex = true`, because that is more in line of what you expect when you write:

```
let foo = [float32 1, 2, 4, 8]
ctx.getType(foo[1])
```

which would return `float32`. 

However, if `typeOfIndex = false` we instead return `array[4, float32]`. 

Currently we only use the case of `typeOfIndex = false`, because I only use this helper to determine the type of a `gpuIndex` of a `gpuDeref` expression, to check if we need to remove the `gpuDeref` layer.

But having a `getType` that works intuitively on `GpuAst` nodes (similar to `getType` on `NimNode`) could be useful. Indeed I had multiple cases where I _almost_ implemented it before, but then realized I didn't quite need it.

constantine/math_compiler/experimental/backends/cuda.nim
Vindaar: What do you have in mind by auto-handle? We _are_ auto handling `HiddenDeref`. What we _might_ want to do in the future is to copy Nim's AST by explicitly differentiating between `gpuDeref` and `gpuHiddenDeref` (which doesn't exist yet).

constantine/math_compiler/experimental/backends/cuda.nim
Vindaar: I have test cases for (I think) all of these (incl the WebGPU ones) as part of the Lita repository. I can move those over here, I suppose.

constantine/math_compiler/experimental/backends/cuda.nim
Vindaar: I updated the explanation.

constantine/math_compiler/experimental/gpu_compiler.nim
Vindaar: Done.

constantine/math_compiler/experimental/backends/cuda.nim
Vindaar: Updated all of these.

constantine/math_compiler/experimental/backends/cuda.nim
mratsim: mutable types are passed as `nnkHiddenDeref` but it doesn't seem like `getInnerPointerType` handles it hence my remark.

```
proc getInnerPointerType(n: NimNode, allowToFail: bool = false, allowArrayIdent: bool = false): GpuType =
  doAssert n.typeKind in {ntyPtr, ntyPointer, ntyUncheckedArray, ntyVar} or n.kind == nnkPtrTy, "But was: " & $n.treerepr & " of typeKind " & $n.typeKind
  if n.typeKind in {ntyPointer, ntyUncheckedArray}:
    let typ = n.getTypeInst()
    doAssert typ.kind == nnkBracketExpr, "No, was: " & $typ.treerepr
    doAssert typ[0].kind in {nnkIdent, nnkSym}
    doAssert typ[0].strVal in ["ptr", "UncheckedArray"]
    result = nimToGpuType(typ[1], allowToFail, allowArrayIdent)
  elif n.kind == nnkPtrTy:
    result = nimToGpuType(n[0], allowToFail, allowArrayIdent)
  elif n.kind == nnkAddr:
    let typ = n.getTypeInst()
    result = getInnerPointerType(typ, allowToFail, allowArrayIdent)
  elif n.kind == nnkVarTy:
    # VarTy
    #   Sym "BigInt"
    result = nimToGpuType(n[0], allowToFail, allowArrayIdent)
  elif n.kind == nnkSym: # symbol of e.g. `ntyVar`
    result = nimToGpuType(n.getTypeInst(), allowToFail, allowArrayIdent)
  else:
    raiseAssert "Found what: " & $n.treerepr
```

mratsim:
The PR is in good shape so merging it to unblock Lita.

Leftover TODOs:

- copy test cases from Lita repos to here for improved maintenance and anti-regression

Vindaar:
> mutable types are passed as nnkHiddenDeref but it doesn't seem like getInnerPointerType handles it hence my remark.

Mutable types are passed as `nnkHiddenDeref`, yes. _But_ when looking at the `typeKind` you get an `ntyVar`.

The `nnkHiddenDeref` appears in the context of the symbol being used inside of a proc. But we deduce the type not based on that, but rather the individual symbols / as part of the parameters of procs / arguments.

It is true though that depending on what node you were to pass into `getInnerPointerType` you wouldn't resolve the hidden addr. But that should never happen. The regular code calls `nimToGpuType`, which handles var types via `ntyVar`. `getInnerPointerType` is only called from there for pointer types.

PR #571, opened 2025-08-29
add: koalabear prime field

agnxsh:
## Koala Bear prime field
$$
2^{31} - 2^{24} + 1
$$


constantine/named/config_fields_and_curves.nim
```diff
@@ -82,7 +73,21 @@ declareCurves:
     testingCurve: true
     bitwidth: 127
     modulus: "0x7fffffffffffffffffffffffffffffff" # 2^127 - 1
+    
   # -----------------------------------------------------------------------------
```
mratsim, line 77: above this are testing "curves" and fields.

PR #574, opened 2025-09-01
allow finite field square root helpers to work with `Fr` as well

Vindaar:
Makes it so that all the finite field procs related to square roots can be run both with `Fp` types as well as `Fr`.

### DIFF
--- a/constantine/math/arithmetic/finite_fields_square_root.nim
+++ b/constantine/math/arithmetic/finite_fields_square_root.nim
@@ -46,7 +46,7 @@ func invsqrt_p3mod4(r: var FF, a: FF) =
   # a^((p-3)/2))        ≡ 1/a  (mod p)
   # a^((p-3)/4))        ≡ 1/√a (mod p)      # Requires p ≡ 3 (mod 4)
   static: doAssert FF.Name.has_P_3mod4_primeModulus()
-  when FF.Name.hasSqrtAddchain():
+  when FF.hasSqrtAddchain():
     r.invsqrt_addchain(a)
   else:
     r = a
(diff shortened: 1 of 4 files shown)


constantine/math/arithmetic/finite_fields_square_root.nim
```diff
@@ -83,7 +83,7 @@ func invsqrt_p5mod8(r: var Fp, a: Fp) =
   #   a^((p-1)/4)) * a⁻¹ ≡  -1/a  (mod p)
   #   a^((p-5)/4))       ≡  -1/a  (mod p)
   #   a^((p-5)/8))       ≡ ± √-1/√a (mod p)
-  # as p ≡ 5 (mod 8), hence 𝑖 ∈ Fp with 𝑖² ≡ −1 (mod p)
+  # as p ≡ 5 (mod 8), hence 𝑖 ∈ FF with 𝑖² ≡ −1 (mod p)
```
mratsim, line 86:
```suggestion
  # as p ≡ 5 (mod 8), hence 𝑖 ∈ Fp with 𝑖² ≡ −1 (mod p)
```

mratsim (commented):
LGTM,

1 nit.

I would expand the tests with some on Fr.

Vindaar:
Yup, will add some tests tomorrow! 

Initially I wanted to add some, but then ran into the trouble that for `Fake103` (which is tested) switching to `Fr` breaks, because it only defines the `modulus` but not the `order` field in the curve definitions.

mratsim:
We can test only on "true" curves.

mratsim:
From https://github.com/mratsim/constantine/pull/574/commits/1e118a6c88e352c75d45b9b6f137f92f59e77403

Adding sqrt support is too complex for actually no use-cases. As discussed in the Groth16 PR, only `isSquare` on Fr is needed. And for the WebGPU small fields codegen, only Fp is needed.

Also in general, when we need sqrt of Fr, it's in the case of an embedded curve like Bandersnatch in BLS12-381, so that's already handled.

In terms of complexity we would need to add all addition chains for r ≡ 3 (mod 4) and r ≡ 5 (mod 8) cases as well as compute Tonelli-Shanks parameters and modify the macro for addchain and Tonelli-Shanks detection. And for Tonelli-Shanks we would need to modify all files to distinguish between Fp and Fr primes.

Furthermore when looked through a binary size profiler (https://github.com/google/bloaty), sqrt is also the biggest code size contributor of Constantine.

Hence closing this PR, and will open one that adds isSquare on Fr.

PR #575, opened 2025-09-01
Remove `importc` in Nvidia wrapped types

Vindaar:
With our current Nvidia wrapper these are actively harmful. With them we get C compiler errors.

Also adds the `cuCtxSetLimit` function, which allows setting custom limits (like increasing the stack size).

### DIFF
--- a/constantine/platforms/abis/nvidia_abi.nim
+++ b/constantine/platforms/abis/nvidia_abi.nim
@@ -825,6 +825,21 @@ type                          ##
                            ##
     CU_JIT_INPUT_NVVM = 5, CU_JIT_NUM_INPUT_TYPES = 6
 
+type
+  CUlimit* {.size: sizeof(cint).} = enum
+    CU_LIMIT_STACK_SIZE                       = 0x00, ## < GPU thread stack size */
+    CU_LIMIT_PRINTF_FIFO_SIZE                 = 0x01, ## < GPU printf FIFO size */
+    CU_LIMIT_MALLOC_HEAP_SIZE                 = 0x02, ## < GPU malloc heap size */
+    CU_LIMIT_DEV_RUNTIME_SYNC_DEPTH           = 0x03, ## < GPU device runtime launch synchronize depth */
+    CU_LIMIT_DEV_RUNTIME_PENDING_LAUNCH_COUNT = 0x04, ## < GPU device runtime pending launch count */
+    CU_LIMIT_MAX_L2_FETCH_GRANULARITY         = 0x05, ## < A value between 0 and 128 that indicates the maximum fetch granularity of L2 (in Bytes). This is a hint */
+    CU_LIMIT_PERSISTING_L2_CACHE_SIZE         = 0x06, ## < A size in bytes for L2 persisting lines cache size */
+    CU_LIMIT_SHMEM_SIZE                       = 0x07, ## < A maximum size in bytes of shared memory available to CUDA kernels on a CIG context. Can only be queried, cannot be set */
+    CU_LIMIT_CIG_ENABLED                      = 0x08, ## < A non-zero value indicates this CUDA context is a CIG-enabled context. Can only be queried, cannot be set */
+    CU_LIMIT_CIG_SHMEM_FALLBACK_ENABLED       = 0x09, ## < When set to zero, CUDA will fail to launch a kernel on a CIG context, instead of using the fallback path, if the kernel uses more shared memory than available */
+    CU_LIMIT_MAX
+
+
 {.pragma: v1, noconv, importc, dynlib: libCuda.}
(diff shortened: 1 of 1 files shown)


mratsim:
I'm curious about your Nim compiler + C compiler + LLVM lib + Cuda version as I don't get any error.

Unless cuda.h is imported somewhere, the compiler has no way to find a pointer mismatch as there are no other definition.

And matching the name is easier when debugging codegen.

Vindaar:
This is with:
- Nim version 2.2.4 (latest stable `f7145dd26efeeeb6eeae6fff649db244d81b212d`)
- GCC version: `14.2.1 20250405`
- or Clang version: `17.0.6`
- CUDA 12.8

and happens on both the LLVM as well as NVRTC test cases with this kind of error:

```
~/src/nim/constantine/nimcache/tests/gpu/t_load_store.nim/@m..@s..@sconstantine@splatforms@sabis@snvidia_abi.nim.c:41:96: error: unknown type name 'CUcontext'
   41 | typedef N_NOCONV_PTR(tyEnum_CUresult__KUCbFqldGRY17wzdY1nAfA, tyProc__t0sUhPdNf3lioKRScnpKgw) (CUcontext* pctx_p0, NU32 flags_p1, NI32 dev_p2);
      |                                                                                                ^~~~~~~~~
~/src/nim/constantine/nimcache/tests/gpu/t_load_store.nim/@m..@s..@sconstantine@splatforms@sabis@snvidia_abi.nim.c:42:98: error: unknown type name 'CUcontext'
   42 | typedef N_NOCONV_PTR(tyEnum_CUresult__KUCbFqldGRY17wzdY1nAfA, tyProc__GmeIR9bOmRNBA9apcjf2wYMw) (CUcontext* module_p0, void* image_p1);
      |                                                                                                  ^~~~~~~~~
~/src/nim/constantine/nimcache/tests/gpu/t_load_store.nim/@m..@s..@sconstantine@splatforms@sabis@snvidia_abi.nim.c:43:96: error: unknown type name 'CUcontext'
   43 | typedef N_NOCONV_PTR(tyEnum_CUresult__KUCbFqldGRY17wzdY1nAfA, tyProc__FBYpOsbG3oqEYhGAzyEkdA) (CUcontext* kernel_p0, CUmodule module_p1, NIM_CHAR* fnName_p2);
      |                                                                                                ^~~~~~~~~
```

I don't see any weird Nim config flags in my configuration either that would explain this. Note that I don't really understand why this happens either. It just seems like when the `importc` is there, it produces function signatures with those types in place, but does not actually include / define the corresponding types in the helper C files. 

This may just be a Nim compiler bug.

mratsim:
It's a regression between nim 2.2.0 and nim 2.2.2

mratsim:
And just to make sure, it's not fixed by adding `nodecl` like in #512 as they are not builtin the compiler.

I'm not sure why it works in my case with Nim 2.2.0 but not in https://github.com/nim-lang/Nim/issues/24604

mratsim:
Minimal repro:

```Nim
const libCuda = "(libcuda.so|cuda.lib)"

{.pragma: v1, noconv, importc, dynlib: libCuda.}
{.pragma: v2, noconv, importc: "$1_v2", dynlib: libCuda.}

type
  CUdevice* = distinct int32
  CUcontext* {.importc.} = distinct pointer

template check(status: cint) =
  let code = status # ensure that the input expression is evaluated once only
  if code != 0:
    writeStackTrace()
    stderr.write(astToStr(status) & " " & $instantiationInfo() & " exited with error: " & $code & '\n')
    quit 1

proc cuInit*(flags: uint32): cint {.v1.}
proc cuDeviceGet*(device: var CUdevice, ordinal: int32): cint {.v1.}
proc cuCtxCreate*(pctx: var CUcontext, flags: uint32, dev: CUdevice): cint {.v2.}

let deviceID = 0'i32
check cuInit(uint32 deviceID)

var cuDevice: CUdevice
check cuDeviceGet(cuDevice, deviceID)

var ctx: CUcontext
check cuCtxCreate(ctx, 0, cuDevice)
```

mratsim:
So distinct void pointer are now using the importc name instead of `void *`

image

mratsim:
Even more minimal:

```Nim
type CUcontext* {.importc.} = distinct pointer

proc foo(ctx: CUcontext) =
  echo "Hello"

var ctx: CUcontext
ctx.foo()
```

mratsim:
Raised upstream https://github.com/nim-lang/Nim/issues/25137

I left the importc to avoid name mangling but instead of CUcontext, we had `void*`

So maybe we actually should use `exportc` but oh well, this might avoid LTO collisions if constantine is build with another Cuda library that imports Nvidia types.

PR #583, opened 2025-10-14
Support more uint literals, simplify registering types for GPU, regular finite field `pow` typo fix

Vindaar:
Decided to put a small typo fix of in-place `pow` for finite fields into this PR, because I noticed it in the same context and it's one character change. 

On the GPU compiler side this:
- adds support for all uint literal types 
- simplifies how we register types in the Nim -> GPU Ast transformation. We now do this in one place only, namely in `nimToGpuType`. We needed to pass the GPU context into this proc, which initially I didn't intend to do. But the code was getting a bit ugly with `maybeAddType` calls all over the place.

### DIFF
--- a/constantine/math/arithmetic/finite_fields.nim
+++ b/constantine/math/arithmetic/finite_fields.nim
@@ -613,7 +613,7 @@ func pow*(r: var FF, a: FF, exponent: BigInt or openArray[byte] or FF) =
   ## ``a``: a field element to be exponentiated
   ## ``exponent``: a finite field element or big integer
   r = a
-  a.pow(exponent)
+  r.pow(exponent)
 
 # Vartime exponentiation
 # -------------------------------------------------------------------
--- a/constantine/math_compiler/experimental/nim_to_gpu.nim
+++ b/constantine/math_compiler/experimental/nim_to_gpu.nim
@@ -11,7 +11,8 @@ import std / [macros, strutils, sequtils, options, sugar, tables, strformat, has
 import ./gpu_types
 import ./backends/backends
 
-proc nimToGpuType(n: NimNode, allowToFail: bool = false, allowArrayIdent: bool = false): GpuType
+proc nimToGpuType(ctx: var GpuContext, n: NimNode, allowToFail: bool = false, allowArrayIdent: bool = false): GpuType
+proc maybeAddType*(ctx: var GpuContext, typ: GpuType)
 
 proc initGpuType(kind: GpuTypeKind): GpuType =
   ## If `kind` is `gtPtr` `to` must be the type we point to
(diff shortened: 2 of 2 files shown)


constantine/math_compiler/experimental/nim_to_gpu.nim
```diff
@@ -1127,14 +1130,18 @@ proc toGpuAst*(ctx: var GpuContext, node: NimNode): GpuAst =
...
+    result.lType = initGpuType(gtUInt64) ## XXX: base on target platform!
+  of nnkUInt16Lit:
     result = GpuAst(kind: gpuLit)
     result.lValue = $node.intVal
-    result.lType = initGpuType(gtUInt64) ## XXX: base on target platform!
+    result.lType = initGpuType(gtUInt16) ## XXX: base on target platform!
+  of nnkUInt32Lit:
+    result = GpuAst(kind: gpuLit)
+    result.lValue = $node.intVal
+    result.lType = initGpuType(gtUInt32) ## XXX: base on target platform!
```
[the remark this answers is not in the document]
Vindaar: Yeah, I know I could bot. :)

constantine/math/arithmetic/finite_fields.nim
```diff
@@ -613,7 +613,7 @@ func pow*(r: var FF, a: FF, exponent: BigInt or openArray[byte] or FF) =
   ## ``a``: a field element to be exponentiated
   ## ``exponent``: a finite field element or big integer
   r = a
-  a.pow(exponent)
+  r.pow(exponent)
```
mratsim, line 616: Crazy that this wasn't tested

constantine/math_compiler/experimental/nim_to_gpu.nim
```diff
@@ -1127,14 +1130,18 @@ proc toGpuAst*(ctx: var GpuContext, node: NimNode): GpuAst =
     result = GpuAst(kind: gpuLit)
     result.lValue = $node.intVal
     result.lType = initGpuType(gtInt32)
-  of nnkUInt32Lit:
+  of nnkUIntLit, nnkUint64Lit:
     result = GpuAst(kind: gpuLit)
     result.lValue = $node.intVal
-    result.lType = initGpuType(gtUInt32)
-  of nnkUIntLit:
+    result.lType = initGpuType(gtUInt64) ## XXX: base on target platform!
```
mratsim, line 1136: uint should probably be mapped to uint32 by default as well on GPU.

constantine/math_compiler/experimental/nim_to_gpu.nim
```diff
@@ -1127,14 +1130,18 @@ proc toGpuAst*(ctx: var GpuContext, node: NimNode): GpuAst =
...
     result.lValue = $node.intVal
-    result.lType = initGpuType(gtUInt64) ## XXX: base on target platform!
+    result.lType = initGpuType(gtUInt16) ## XXX: base on target platform!
+  of nnkUInt32Lit:
+    result = GpuAst(kind: gpuLit)
+    result.lValue = $node.intVal
+    result.lType = initGpuType(gtUInt32) ## XXX: base on target platform!
   of nnkFloat64Lit, nnkFloatLit:
     result = GpuAst(kind: gpuLit)
     result.lValue = $node.floatVal & "f"
```
mratsim, line 1147: iirc, f is float32, and default float in C/Metal/Cuda is float32 not float64 iirc

mratsim (approved):
LGTM apart from the default value which should be 32-bit for int, uint float I think.

Vindaar:
While I agree in principle about the type mappings, in practice I'm starting to consider the opposite approach. Because the code is part of a typed macro and we want the Nim compile time guarantees, if we map int, uint, float to different types than in regular Nim, we deal with many very annoying type mismatch bugs on the Nim side. In the end one anyways has to to all sorts of manual type conversions. Say if int is int32 then it should be fine to add an int to an int32 variable, or assign int to int32. but of course that's not valid Nim.

So I'm considering to keep Nim semantics (which this PR of course doesn't do yet) and document that.

mratsim:
Let's merge it in like that for now then.

I do think it would be quite verbose to always read u32 / f32 in GPU code but:
- we probably would write generic algorithms most of the time and leave instantiation type to the caller
- we can create converters template pragma, for example `{.conv32.}` that signals that all types are 32-bit within the proc scope.

PR #589, opened 2026-03-15
[EC] Fix mutable assignment in batch affine - followup #579

mratsim:
Currently we have a failure `constantine/constantine/math/elliptic/ec_shortweierstrass_batch_ops.nim(52, 3) Error: 'SecretBool(affs[0].y.mres.limbs[0])' cannot be assigned to`

This is because after #579, the `SecretWord myExpression` creates an immutable expression

https://github.com/mratsim/constantine/blob/cedcf0da4b72964154edd1701bb51ae77426e6c5/constantine/math/elliptic/ec_shortweierstrass_batch_ops.nim#L45-L59

The strange thing is that, when compiling through `nimble test` or `nimble test_parallel` that do a simple exec, compilation works and test is run:

image

but compiling the code directly triggers a compiler error.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced internal consistency and performance in cryptographic batch operations through improved type handling in elliptic curve computations.

### DIFF
--- a/constantine/math/elliptic/ec_shortweierstrass_batch_ops.nim
+++ b/constantine/math/elliptic/ec_shortweierstrass_batch_ops.nim
@@ -42,21 +42,21 @@ func batchAffine*[F, G](
 
   # To avoid temporaries, we store partial accumulations
   # in affs[i].x and affs[i].y will store if the input was 0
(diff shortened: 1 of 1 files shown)


mratsim:
@greptile

PR #598, opened 2026-04-20
Workaround GMP regressions and breaking changes from https://github.com/subsetpark/nim-gmp/pull/3

mratsim:
This workarounds 2 breaking changes from upstream nim-gmp that is used for fuzzing:

- `csize` (alias for int in Nim) -> `csize_t` which maps exactly to C csize_t, unfortunately there is no version tagged so we might want to pin an exact commit in the future to avoid such regression now that nimble supports lockfiles. But I'm unsure if optinal dependencies locking actual works
-  https://github.com/subsetpark/nim-gmp/pull/3 reintroduced destructors in a low-level API despite me removing them in https://github.com/subsetpark/nim-gmp/pull/2 and that broke usage of mpz_clear ... again.

## Summary by CodeRabbit

* **Tests**
  * Updated GMP function calls across benchmarks and test suites with explicit C-type conversions for improved type safety.
  * Adjusted memory cleanup patterns in GMP object handling throughout test files.
  * Modified type handling for size and count parameters in cryptographic operation tests.

* **Chores**
  * Updated benchmark code to use explicit typed arguments in library function calls.

### DIFF
--- a/benchmarks/bench_gmp_modexp.nim
+++ b/benchmarks/bench_gmp_modexp.nim
@@ -83,10 +83,10 @@ for i in 0 ..< 5:
     mpz_init(mm)
     mpz_init(rr)
 
-    aa.mpz_import(a.limbs.len, GMP_LeastSignificantWordFirst, sizeof(SecretWord), GMP_WordNativeEndian, 0, a.limbs[0].unsafeAddr)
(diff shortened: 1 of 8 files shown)


mratsim:
@coderabbitai review

Issue #522, opened 2025-01-17
Benchmarks UI improvement: Markdown Table output


mratsim:
Currently the benchmarks require a screenshot to be shared or formatting might be messed up with line wrap or too long lines.

It would be nice to report them in a format that could be copy-pasted directly to Github and ethresear.ch and ethereum-magicians.org so we should change the output to a Markdown table output.

PR #621, opened 2026-06-18
Render summary benchmark results as Markdown tables

reflecttypefor:
## Summary

This adds a small Markdown table reporting helper for benchmark output and uses it in the summary benchmark template.

The summary benchmark output can now be copied directly into GitHub, ethresear.ch, and ethereum-magicians.org without losing alignment due to line wrapping.

This is a first focused step toward #522, keeping the change limited to the summary benchmarks before expanding the same helper to other benchmark suites.

## Changes

- Add `benchmarks/bench_report.nim` with Markdown table formatting helpers
- Escape Markdown table cells containing pipes or newlines
- Render `bench_summary_template` benchmark rows as Markdown tables
- Add tests for Markdown header/separator/row formatting
- Register the new test in `constantine.nimble`

## Summary by CodeRabbit

* **New Features**
  * Added Markdown table formatting for benchmark reports with automatic cell escaping and proper table structure generation.

* **Tests**
  * Added test suite for report formatting functionality to verify Markdown table rendering and cell escaping.

* **Chores**
  * Updated build configuration to include new test suite.


mratsim:
Thank you for the PR.

I think the next step would be to modify 2 benchmark files as a branch that build on this branch and a "staggered" PR that targets this PR branch so we can see what it looks like.

I suggest: 
- https://github.com/mratsim/constantine/blob/master/benchmarks/bench_fp.nim
- https://github.com/mratsim/constantine/blob/master/benchmarks/bench_summary_bls12_381.nim

as they are very standardized while some are somewhat customized and would need more work (https://github.com/mratsim/constantine/blob/master/benchmarks/bench_eth_evm_precompiles.nim)

You can run them with `CC=clang nimble bench_fp`, `CC=clang nimble bench_summary_bls12_381` and in general the benchmarks are added to a nimble task in https://github.com/mratsim/constantine/blob/master/constantine.nimble#L1034-L1275

PR #623, opened 2026-07-14
Fix threadpool hangs with `numThreads >= 257`

nitely:
Fix threadpool hangs with `numThreads >= 257`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved threadpool event handling for configurations with large numbers of waiting threads.
  * Prevented potential hangs when queuing many tasks and shutting down the threadpool.

* **Tests**
  * Added stress coverage for threadpools with 257 threads and multiple queued tasks.


mratsim:
Good catch

PR #626, opened 2026-07-29
[PeerDAS] Expose `compute_cells` (cells without proofs) through the C ABI and constantine-go

syjn99:
On the PeerDAS/Fulu block proposal hot path, consensus clients receive cell **proofs** from the execution layer for free (`BlobsBundleV2` / `engine_getBlobsV2`) and only need the erasure-extended **cells**.

I'm assessing Constantine for Prysm's KZG operations (OffchainLabs/prysm#16850); in our benchmarks of constantine-go against c-kzg-4844 Constantine wins every measured operation, but this API gap is currently the main blocker to adoption: Prysm calls a cells-only `ComputeCells` at

- `beacon-chain/blockchain/kzg/validation.go` (proof verification path), and
- `beacon-chain/core/peerdas/reconstruction.go` (data-column reconstruction)

c-kzg-4844, go-eth-kzg and rust-eth-kzg all expose the equivalent entry point, matching the consensus-spec function [`compute_cells`](https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/polynomial-commitments-sampling.md#compute_cells). With this change, migrating to constantine from other kzg libraries would be much easier.

## Summary by CodeRabbit

* **New Features**
  * Added support for computing all 128 EIP-7594 cells from an extended blob without generating KZG proofs.
  * Exposed cell computation through the C and Go APIs, including error handling for invalid inputs.

* **Tests**
  * Added Go test-vector coverage for cells-only computation.
  * Updated PeerDAS benchmarks and tests for the new cell-computation interface.

### DIFF
--- a/benchmarks/bench_eth_eip7594_peerdas.nim
+++ b/benchmarks/bench_eth_eip7594_peerdas.nim
@@ -173,7 +173,7 @@ proc benchComputeCells(b: BenchSet, ctx: ptr EthereumKZGContext, iters: int) =
   var cells: ref array[CELLS_PER_EXT_BLOB, Cell]
   new(cells)
   bench("compute_cells (half-FFT optimization)", iters):
-    doAssert cttEthKzg_Success == ctx.compute_cells(cells[], b.blobs[0])
+    doAssert cttEthKzg_Success == ctx.compute_cells(cells[].asUnchecked(), b.blobs[0])
 
 proc benchComputeCellsAndKZGProofsNoPrecomp(b: BenchSet, ctx: ptr EthereumKZGContext, iters: int) =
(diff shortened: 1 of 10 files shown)


mratsim:
SGTM, will create a separate issue to expose the same in Rust as well as this is go only.

Seems like there are unrelated CI stuff to fix though. Will have a look

mratsim:
Hopefully CI is unblocked by #627

syjn99:
@mratsim thanks, will wait until #627 got merged.

mratsim:
The go tests are done first and are passing. The issue left seems to be a regression in an unreleased Nim v2.2.x that affects unit tests but not integration tests or building the static library so merging. Regression tracked in #628.

